PR #8, opened 2025-06-18
[AIT-238] fix charts and KPIs

natacha-scopeo:


jccscopeo (approved):
Working locally, I think for a first version for them it is enough. Maybe in the future we will allow more granularity of the time period to select

jccscopeo (approved):
I think we could try to add more explanations for the graphics, and maybe remove the %nan or %0 of the stats, otherwise of a first version, we have we need to need for the ACPM (usage per day)

dianascopeo (approved):
Worked locally. Would be better to have x and y axis labels for the next version :)

dianascopeo (approved):

PR #15, opened 2025-06-20
[AIT-235] fix when max usage reach and api key is provided

natacha-scopeo:


ada_backend/database/seed_project_db.py
```diff
@@ -226,5 +228,11 @@ def seed_projects_db(session: Session):
 
     logging.basicConfig(level=logging.INFO)
 
+    set_trace_manager(TraceManager(project_name="ada_backend"))
+    trace_manager = get_trace_manager()
+    trace_manager.organization_id = DEFAULT_ORGANIZATION_ID
+    trace_manager.project_id = PROJECT_UUIDS["project_1"]
+    trace_manager.organization_llm_providers = ["openai", "anthropic", "cohere"]
+
```
dianascopeo, line 236: I think we are not running this file now and we cannot have this here

jccscopeo (commented):
Capture d’écran 2025-06-20 à 15 04 07
I have this error when the number of tokens is exceeded. I did trace-db-upgrade but should I delete it first?

engine/llm_services/llm_service.py
```diff
@@ -1,33 +1,52 @@
...
+        provider: str,
     ):
         self.trace_manager = trace_manager
         self._completion_model: str = None
         self._embedding_model: str = None
         self._default_temperature: float = None
+        self.provider = provider
 
     @abc.abstractmethod
+    @with_usage_check
```
jccscopeo, line 41: Do we count embedding token size the same a chat? i.e maybe testing some tiny pdfs will still explode the user token size of the organization. But we can keep like this for now

jccscopeo (approved):
Working locally. The wrapper is a good idea, we should try to see how we can increase the token size depending on what services are used (i.e how to count token for embedding, is it the same as chat etc)

jccscopeo (approved):
Working locally with an AI agent + internet search tool

jccscopeo (approved):
Working locally with AI agent + RAG and Internet Search tool. The ingestion is failing when using llm (for pdf).

PR #16, opened 2025-06-20
Refacto llm service

marc-scopeo:


data_ingestion/document/document_chunking.py
```diff
@@ -60,28 +59,29 @@ def get_chunks_dataframe_from_doc(
 ) -> pd.DataFrame:
     all_chunks = []
     LOGGER.info(f"Processing document {document.file_name} of type {document.type}")
-    description_doc = None
-    if add_doc_description_to_chunks:
-        if documents_summary_func is None or add_summary_in_chunks_func is None:
-            raise ValueError("No summary function / put summary function provided for documents")
-        description_doc = documents_summary_func(document)
+    # TODO: add summary to chunks when we will have a gemini functional service
+    # description_doc = None
```
natacha-scopeo, line 63: Why is it commit ?

dianascopeo: we are not using that part and it is using gemini code which we dont have for now thats why commented and add a TODO

natacha-scopeo (approved):

jccscopeo (approved):
Working locally with ingestion, RAG, AI agent , LLM call agent, vocabulary search, internet with openai

PR #28, opened 2025-06-24
[AIT-265] update db form

natacha-scopeo:


ingestion_script/ingest_db_source.py
```diff
@@ -26,17 +26,25 @@
...
+        DBColumn(name=chunk_column_name, type="VARCHAR"),
     ]
     if timestamp_column_name:
         columns.append(DBColumn(name=timestamp_column_name, type="VARCHAR"))
 
+    if url_column_name:
+        columns.append(DBColumn(name=url_column_name, type="VARCHAR"))
+
     if metadata_column_names:
         columns.extend(DBColumn(name=col, type="VARCHAR") for col in metadata_column_names)
```
dianascopeo, line 49: for metadata we will not have same as drive to have metadata and then dict?

dianascopeo: Capture d’écran 2025-06-26 à 13 29 43

natacha-scopeo: no i was one column for each metadata

dianascopeo: I think we should have same format as pdf files

dianascopeo (approved):
worked locally using snowflake db

jccscopeo (approved):
Working locally

PR #35, opened 2025-06-30
Add python code interpreter with E2B

marc-scopeo:


dianascopeo (approved):
Worked locally and plots in notebook

PR #36, opened 2025-07-01
[AIT-259] remove open ai telemetry

natacha-scopeo:


jccscopeo (approved):
Working for AI agent with RAG and internet search. Retriever does not display embedding model though.

PR #40, opened 2025-07-02
Bugfix : make a function to handle images in react function calling i…

marc-scopeo:
…nstead of try except


engine/agent/react_function_calling.py
```diff
@@ -249,3 +245,19 @@ def get_dummy_ai_agent_description() -> ToolDescription:
...
+def get_images_from_message(messages: list[ChatMessage]) -> list[str]:
+    if messages:
+        message = messages[-1]
+        if "}" in message.content:
+            json_end = message.content.rfind("}") + 1
+            try:
+                json_content = json.loads(message.content[:json_end])
+                imgs = [result["png"] for result in json_content["results"] if "png" in result]
+            except (json.JSONDecodeError, TypeError):
+                LOGGER.debug("No images found in the response.")
```
natacha-scopeo, line 260: we only logger no image in the try except ? not for all the function ?

natacha-scopeo (approved):

PR #54, opened 2025-07-07
AIT-328-fix excel and csv ingetsion

dianascopeo:


jccscopeo (commented):
Do we need to also put the overlapping parameters as an argument for the ingestion endpoint and functions?

dianascopeo:
> Do we need to also put the overlapping parameters as an argument for the ingestion endpoint and functions?

i dont have overlap for excel and csv files

natacha-scopeo (approved):

PR #63, opened 2025-07-10
feat(Code robustness): Allow all the agents and tool (except sql or database related tools) to be async.

jccscopeo:
- Allow LLMService to have async methods
- Allow Qdrant service to have async methods
- Allow reranker (cohere) to be async
- Change React, RAG, API Tool, python interpreter to be async (propagate async changes on sub components such as retriever, synthesizer etc)
- The code does not transform database related services to be async


tests/agent/test_react_function_calling.py
```diff
@@ -92,28 +96,168 @@ def test_run_with_tool_calls(agent_calls_mock, get_span_mock, react_agent, agent
...
+            self.id = "1"
+            self.function = SimpleNamespace(name="test_tool", arguments=json.dumps({"test_property": "Test value"}))
+            self.type = "function"
+
+        def model_dump(self):
+            return {
+                "id": "1",
+                "function": {"name": "test_tool", "arguments": json.dumps({"test_property": "Test value"})},
+                "type": "function",
+            }
```
pabloscopeo, line 190: Maybe we can reuse this? It could be on one of the mock files (or its own).

tests/agent/test_react_function_calling.py
```diff
@@ -92,28 +96,168 @@ def test_run_with_tool_calls(agent_calls_mock, get_span_mock, react_agent, agent
...
+            "content": "Tool response",
+            "tool_calls": [
+                {
+                    "id": "1",
+                    "function": {"name": "test_tool", "arguments": json.dumps({"test_property": "Test value"})},
+                    "type": "function",
+                }
+            ],
+        },
+    )
```
pabloscopeo, line 210: Same for this?

tests/agent/test_react_function_calling.py
```diff
@@ -132,34 +276,64 @@ def test_max_iterations(agent_calls_mock, get_span_mock, react_agent, agent_inpu
...
+            self.id = "1"
+            self.function = SimpleNamespace(name="test_tool", arguments=json.dumps({"test_property": "Test value"}))
+            self.type = "function"
+
+        def model_dump(self):
+            return {
+                "id": "1",
+                "function": {"name": "test_tool", "arguments": json.dumps({"test_property": "Test value"})},
+                "type": "function",
+            }
```
pabloscopeo, line 292: Idem

tests/agent/test_react_function_calling.py
```diff
@@ -132,34 +276,64 @@ def test_max_iterations(agent_calls_mock, get_span_mock, react_agent, agent_inpu
...
+            "1": AgentPayload(messages=[ChatMessage(role="assistant", content="Tool response")], is_final=False)
+        }
+        processed_tool_calls = [
+            {
+                "id": "1",
+                "function": {"name": "test_tool", "arguments": json.dumps({"test_property": "Test value"})},
+                "type": "function",
+            }
+        ]
+        return tool_outputs, processed_tool_calls
```
pabloscopeo, line 333: Idem

tests/llm_services/test_llm_service.py
```diff
@@ -1,20 +1,33 @@
...
 
+@pytest.fixture(autouse=True)
+def setup_tracing_context():
+    """Set up tracing context for all tests in this module."""
+    set_tracing_span(
+        project_id="test-project",
+        organization_id="test-org",
+        organization_llm_providers=["openai", "mistral", "google", "cohere"],
+        conversation_id="test-conversation",
+    )
```
pabloscopeo, line 24: I think this conflicts with the refactor done in this [PR](https://github.com/Scopeo/draftnrun/pull/57), but not sure. In general (except for specific integration tests like `test_monitor_endpoint.py`), I think it would be better to mock/disable the tracing system, as it's not what's being tested.

tests/llm_services/test_llm_service.py
```diff
@@ -25,8 +38,23 @@ def test_completion_service():
     assert len(response) > 0
 
 
+@pytest.mark.asyncio
```
pabloscopeo, line 41: It's there a reason to run the tests as async? If it's only for being able to run async fonctions inside, we could just do `asyncio.run(async_function())`.

tests/qdrant/test_qdrant_service.py
```diff
@@ -133,6 +133,129 @@ def test_qdrant_service():
...
+    ]
+    qdrant_agentic_service = QdrantService.from_defaults(
+        embedding_service=embedding_service,
+        default_collection_schema=qdrant_schema,
+    )
+
+    # Ensure a clean state
+    if await qdrant_agentic_service.acollection_exists(TEST_COLLECTION_NAME):
+        await qdrant_agentic_service.adelete_collection(TEST_COLLECTION_NAME)
+    assert not await qdrant_agentic_service.acollection_exists(TEST_COLLECTION_NAME)
```
pabloscopeo, line 176: Is this a mock collection or a real one? Would that mean that we will need to run Qdrant in local for local tests?

pabloscopeo (commented):
Left some comments!

In general, I have two questions:

- Do we really need to make some tests async? I think it adds complexity to the testing if it's just for being able to execute async functions. In that case, we can probably just use `run(async_function())`.

- I'm not a big fan of prefixing "a" to functions to indicate that they're async. I think it just makes functions a bit harder to read (and also a bit confusing in some cases, like `aadd_`). This is more a matter of style, but considering that this is the first async refactor in the codebase, it's the perfect moment to define a convention here. IMO, linters and IDEs are enough to show what's async and what isn’t, so we probably don't need to prefix it.

tests/qdrant/test_qdrant_service.py
jccscopeo: we already run qdrant in test with cloud version I think

tests/llm_services/test_llm_service.py
jccscopeo: You are right. We will fix the test after the merge of the other PR.  On the other hand we don t have test for the trace I think so it is not bad I think to have tests with it on the llm service

tests/llm_services/test_llm_service.py
jccscopeo: As I propagate the async on every other tests, I would like to keep it like that if it is ok

tests/qdrant/test_qdrant_service.py
pabloscopeo: ah ok. We should probably organize our tests better so we know which ones are integration tests and which ones aren't. I would have thought that [test_qdrant_service.py](https://github.com/Scopeo/draftnrun/pull/63/files#diff-71838e58cba906b0287a88fa194480a7c4c79127e146fa4d0cc6edfeb68c8ba2)'s  point was to test the logic, so maybe we need to mock all the rest (but probably not now)

PR #69, opened 2025-07-16
Ait 381 full coding agent

marc-scopeo:


engine/agent/react_function_calling.py
```diff
@@ -236,16 +258,19 @@ async def _run_without_trace(self, *inputs: AgentPayload | dict, **kwargs) -> Ag
             self._current_iteration += 1
             return await self._run_without_trace(agent_input)
         else:  # This should not happen if the "tool_choice" parameter works correctly on the LLM service
-            LOGGER.error(
-                f"Reached the maximum number of iterations ({self._max_iterations}) and still asks for tools."
-                " This should not happen."
+            self.log_trace_event(
+                message=(
+                    f"Reached the maximum number of iterations ({self._max_iterations}). "
+                    f"Returning the fallback answer: {DEFAULT_FALLBACK_REACT_ANSWER}"
```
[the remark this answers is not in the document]
pabloscopeo: Yes, I agree, I think trace events are shorter

CLAUDE.md
pabloscopeo, line 1: I agree with the idea but this is one more doc to maintain. Considering that all this info it's on Readme, is it mandatory?

ada_backend/database/seed/seed_terminal_command_e2b_tool.py
```diff
@@ -0,0 +1,54 @@
...
+        function_callable=True,
+        can_use_function_calling=False,
+        release_stage=db.ReleaseStage.BETA,
+        default_tool_description_id=TOOL_DESCRIPTION_UUIDS["terminal_command_e2b_tool_description"],
+    )
+    upsert_components(session, [terminal_command_e2b_component])
+
+    terminal_command_e2b_parameter_definitions = [
+        ComponentParameterDefinition(
+            id=UUID("e2b10001-1111-2222-3333-444444444444"),
```
pabloscopeo, line 35: it's this something we are doing (using uuids like this)? maybe I missed it

engine/agent/agent.py
```diff
@@ -183,7 +197,7 @@ async def run(
                     {
                         SpanAttributes.TOOL_NAME: self.tool_description.name,
                         SpanAttributes.TOOL_DESCRIPTION: self.tool_description.description,
-                        SpanAttributes.TOOL_PARAMETERS: json.dumps(kwargs),
+                        SpanAttributes.TOOL_PARAMETERS: json.dumps(_filter_serializable_kwargs(kwargs)),
```
pabloscopeo, line 200: Isn't this doing `json.dumps(json.dumps(x))`? I get this is meant for flat dicts in which case it's ok, but are we sure it's only for flat dicts?
Ex: `kwargs={"some_arg": {"inner_arg": value}}`
if `value` it's not JSON serializable, we would have ` f"<dict object>"` instead of `"{"some_arg": {"inner_arg": <type-of-value>}}"`, loosing some info.

engine/agent/react_function_calling.py
```diff
@@ -67,6 +69,26 @@ def __init__(
...
+
+    def _cleanup_shared_sandbox(self) -> None:
+        """Safely clean up the shared sandbox."""
+        if self._shared_sandbox:
+            try:
+                self._shared_sandbox.kill()
+            except Exception as e:
+                LOGGER.error(f"Failed to kill shared sandbox: {e}")
+            finally:
+                self._shared_sandbox = None
```
pabloscopeo, line 91: We are adding this logic directly to ReAct agent which is our base agent practically. It's that ok? Is there a reason to not have a particular agent/tool for this?

engine/agent/react_function_calling.py
```diff
@@ -79,7 +101,13 @@ async def _run_tool_call(
...
             tool for tool in self.agent_tools if tool.tool_description.name == tool_function_name
         )
-        tool_output = await tool_to_use.run(*original_agent_inputs, **tool_arguments)
+        if tool_function_name in CODE_RUNNER_TOOLS:
+            tool_arguments["shared_sandbox"] = self._ensure_shared_sandbox()
+        try:
+            tool_output = await tool_to_use.run(*original_agent_inputs, **tool_arguments)
+        except Exception as e:
+            LOGGER.error(f"Error running tool {tool_function_name}: {e}")
+            tool_output = e
```
pabloscopeo, line 110: Maybe this is the only way to do it, but I feel like we can't have if statement on the base agent for every tool. Unless this is a really special/particular use case, I'd prefer this logic to go on it's own agent.

engine/agent/tools/python_code_interpreter_e2b_tool.py
```diff
@@ -59,13 +59,11 @@ def _extract_images_from_results(self, execution_result: dict) -> list[str]:
...
-                # E2B automatically detects matplotlib plots and adds them as base64 encoded images
-                for image_format in ["png", "jpg", "jpeg", "svg"]:
-                    if image_format in result and result[image_format]:
-                        # The image is already base64 encoded according to E2B docs
-                        images.append(result[image_format])
+            for image_format in ["png", "jpg", "jpeg", "svg", "gif", "webp"]:
+                if hasattr(result, image_format):
+                    image_data = getattr(result, image_format)
+                    if image_data:
+                        images.append(image_data)
```
pabloscopeo, line 66: Nitpick: A bit too nested. Maybe `for image_format not in [...]: continue` could help?

engine/agent/tools/python_code_interpreter_e2b_tool.py
```diff
@@ -75,34 +73,76 @@ def execute_python_code(self, python_code: str, shared_sandbox: Optional[Sandbox
...
+                    if v is None:
+                        continue
+                    elif isinstance(v, (str, int, float, bool, list, dict)):
+                        result_dict[k] = v
+                    else:
+                        try:
+                            result_dict[k] = str(v)
+                        except Exception:
+                            result_dict[k] = f"<{type(v).__name__} object>"
+                serializable_results.append(result_dict)
```
pabloscopeo, line 136: We could have a function for serialization, and use it here and on the other place (`_filter_serializable_kwargs`)

ada_backend/database/seed/seed_terminal_command_e2b_tool.py
marc-scopeo: sonnet 4 decided so... haha
What is important is to have them different from each other right ? Or do you see more problems ?

pabloscopeo: Nah I'm okay with it if it's only for seed data

pabloscopeo:
I think we need to document the E2B API Key somewhere, at least in `.env.example`
image

pabloscopeo (approved):
Code is functional!

image

It would be great if we could address some of the raised points before merging.
- A bit of code dup on serialization
- Putting quick TODOs just so others don't continue adding too much tool logic on ReActAgent
- Document `E2B API Key` somewhere.

Also, while the code it's functional, it generated fake data when using python. Maybe we could force the prompt to not do this unless explicitly asked to.

engine/trace/serializer.py
```diff
@@ -0,0 +1,62 @@
+import json
+from typing import Any, Optional
+
+
+def serialize_to_json(obj: Any, indent: int = 2) -> str:
```
pabloscopeo, line 7: Maybe we shouldn't put `indent = 2` by default. Probably, most of the time, we want to send serialized data in a compact way, so it's faster to send and lighter to store. I think it's better to prettify it the moment it's needed.
We could replace `indent` by `**kwargs` and call `json.dumps(serialized_obj, **kwargs)`, so if needed, someone could still call `serialize_to_json(some_obj, indent=2)`.

engine/agent/react_function_calling.py
```diff
@@ -68,10 +73,31 @@ def __init__(
         self._completion_service = completion_service
         self.input_data_field_for_messages_history = input_data_field_for_messages_history
         self._allow_tool_shortcuts = allow_tool_shortcuts
+        self._shared_sandbox: Optional[AsyncSandbox] = None
+        self._e2b_api_key = getattr(settings, "E2B_API_KEY", None)
```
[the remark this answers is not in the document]
marc-scopeo: in case it is not defined

engine/trace/sql_exporter.py
```diff
@@ -84,12 +84,12 @@ def export(self, spans: list[ReadableSpan]) -> SpanExportResult:
         for span in spans:
             cumulative_error_count = int(span.status.status_code is StatusCode.ERROR)
             try:
-                cumulative_llm_token_count_prompt = int(span.attributes.get(SpanAttributes.LLM_TOKEN_COUNT_PROMPT))
+                cumulative_llm_token_count_prompt = int(span.attributes.get(SpanAttributes.LLM_TOKEN_COUNT_PROMPT, 0))
             except BaseException:
```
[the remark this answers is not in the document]
marc-scopeo: yes. If this was intended, we need to talk

PR #71, opened 2025-07-16
feat(BugFix Trace): Fix the span export problem

jccscopeo:
Fix the span export problem when exporting a bad span leads to a sqlalchemy connection stuck into a stale state:

- allow rollback when having span export errors
- improve logging of errors with span exports
- improve handling of errors/edge cases with spans + better logging
- Add integration tests for the trace

### DIFF
--- a/engine/trace/sql_exporter.py
+++ b/engine/trace/sql_exporter.py
@@ -63,30 +63,22 @@ class SQLSpanExporter(SpanExporter):
 
     def get_org_info_from_ancestors(self, parent_id: str) -> tuple[str, str] | None:
         """Get org_id and org_llm_providers from ancestors of the span."""
-        try:
-            while parent_id:
+        while parent_id:
+            row = self.session.execute(
+                select(models.Span.attributes, models.Span.parent_id).where(models.Span.span_id == parent_id)
+            ).first()
+            if not row:
+                break
+            attributes, parent_id = row
+            if attributes:
                 try:
-                    row = self.session.execute(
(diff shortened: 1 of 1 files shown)


engine/trace/sql_exporter.py
```diff
@@ -57,137 +59,324 @@ def convert_to_list(obj: Any) -> list[str] | None:
 
 class SQLSpanExporter(SpanExporter):
     def __init__(self):
-        self.session = get_session_trace()
+        # Don't create session here - create fresh session for each export
+        self.engine = None
+        self.Session = None
+        self._setup_engine()
```
dianascopeo, line 65: why we need to create fresh session for each export? I dont get if this will solve the bug that we saw

engine/trace/sql_exporter.py
```diff
@@ -61,133 +63,281 @@ def __init__(self):
...
         """Get org_id and org_llm_providers from ancestors of the span."""
-        while parent_id:
-            row = self.session.execute(
-                select(models.Span.attributes, models.Span.parent_id).where(models.Span.span_id == parent_id)
-            ).first()
-            if not row:
-                break
-            attributes, parent_id = row
-            if attributes:
+        try:
```
natacha-scopeo, line 66: why 2 try nested?

jccscopeo: you are right, it is useless. It was to catch some errors but it is useless in this function. I revert to the previous version

PR #74, opened 2025-07-16
AIT-334-add query filter for db ingestion

dianascopeo:


ingestion_script/ingest_folder_source.py
```diff
@@ -76,12 +76,13 @@ def sync_chunks_to_qdrant(
     collection_name: str,
     db_service: DBService,
     qdrant_service: QdrantService,
+    query_filter: Optional[str] = "",
```
natacha-scopeo, line 79: idem to None

ingestion_script/main.py
```diff
@@ -16,6 +16,7 @@
 DEFAULT_CHUNK_SIZE = 1024
 DEFAULT_CHUNK_OVERLAP = 0
 DEFAULT_REPLACE_EXISTING = False
+DEFAULT_QUERY_FILTER = ""
```
natacha-scopeo, line 19: default to None ? so we don't need to add it here

ingestion_script/main.py
```diff
@@ -55,6 +56,9 @@ def ingestion_main(
     replace_existing = source_attributes.get("replace_existing")
     if replace_existing is None:
         replace_existing = DEFAULT_REPLACE_EXISTING
+    query_filter = source_attributes.get("query_filter")
+    if query_filter is None:
+        query_filter = DEFAULT_QUERY_FILTER
```
natacha-scopeo, line 61: to remove as we want default to be None

engine/qdrant_service.py
```diff
@@ -697,21 +701,36 @@ def get_collection_data(self, collection_name: str) -> pd.DataFrame:
                     row[field] = payload.get(field)
             rows.append(row)
 
-        return pd.DataFrame(rows)
+        df = pd.DataFrame(rows)
+        if query_filter:
+            import re
```
natacha-scopeo, line 706: Why import into the function?

engine/qdrant_service.py
```diff
@@ -1056,13 +1052,20 @@ def get_collection_data(self, collection_name: str) -> pd.DataFrame:
...
+
+            converted_query_filter = re.sub(r"(?<![!><])=(?!=)", "==", query_filter)
```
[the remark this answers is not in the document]
dianascopeo: yes maybe version 2 we will implement that

engine/qdrant_service.py
```diff
@@ -1159,11 +1175,28 @@ def sync_df_with_collection(self, df: pd.DataFrame, collection_name: str) -> boo
             LOGGER.info(f"Sync successful : number of points in Qdrant is {n_points}")
             return True
 
-    async def sync_df_with_collection_async(self, df: pd.DataFrame, collection_name: str) -> bool:
-        """
-        Async version of sync_df_with_collection.
-        Synchronize a DataFrame with a Qdrant collection asynchronously.
-        """
+    async def sync_df_with_collection_async(
```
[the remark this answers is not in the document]
natacha-scopeo: do we have a sync function of this async function ? I think Pablo's suggestion not to add async if we don't have a sync function is a good idea.

ingestion_script/ingest_db_source.py
```diff
@@ -163,6 +169,7 @@ def upload_db_source(
         collection_name=qdrant_collection_name,
         db_service=db_service,
         qdrant_service=qdrant_service,
+        query_filter=None,  # Don't apply filter to ingestion DB - data is already filtered
```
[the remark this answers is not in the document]
dianascopeo: for me sync with date I will use their date column and say everything > 2025 I will do sync and the other ones I will keep as it is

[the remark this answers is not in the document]
dianascopeo: what do you mean?

engine/storage_service/db_service.py
```diff
@@ -104,21 +111,42 @@ def update_table(
...
                 if timestamp_column_name
-                else f"SELECT {id_column_name} FROM {target_table_name};"
+                else f"SELECT {id_column_name} FROM {target_table_name}"
             )
+            if query_filter:
+                query += f" WHERE {query_filter}"
+            if timestamp_filter:
+                if query_filter:
+                    query += f" AND {timestamp_column_name} {timestamp_filter}"
+                else:
```
natacha-scopeo, line 125: I this you can simplify this code. if, elif,

engine/storage_service/db_service.py
```diff
@@ -104,21 +111,42 @@ def update_table(
...
+                    query += f" WHERE {timestamp_column_name} {timestamp_filter}"
+            query += ";"
             old_df = self._fetch_sql_query_as_dataframe(query)
             old_df = convert_to_correct_pandas_type(old_df, id_column_name, table_definition)
 
+            # Separate query for existence check (without timestamp filter)
+            existence_query = f"SELECT {id_column_name} FROM {target_table_name}"
```
natacha-scopeo, line 132: what is existence query ?

engine/storage_service/db_service.py
```diff
@@ -104,21 +111,42 @@ def update_table(
...
+            existing_ids_df = convert_to_correct_pandas_type(existing_ids_df, id_column_name, table_definition)
+
             common_df = new_df.merge(old_df, on=id_column_name, how="inner")
             if timestamp_column_name:
-                ids_to_update = set(
-                    common_df[common_df[timestamp_column_name + "_x"] > common_df[timestamp_column_name + "_y"]][
-                        id_column_name
-                    ]
-                )
+                if timestamp_filter:
```
natacha-scopeo, line 141: really hard to read with the 2 if

engine/storage_service/db_service.py
```diff
@@ -104,21 +111,42 @@ def update_table(
...
+                    ids_to_update = set(common_df[id_column_name])
+                else:
+                    ids_to_update = set(
+                        common_df[common_df[timestamp_column_name + "_x"] > common_df[timestamp_column_name + "_y"]][
+                            id_column_name
+                        ]
+                    )
+                print("ids_to_update", ids_to_update)
```
natacha-scopeo, line 149: print

engine/storage_service/local_service.py
```diff
@@ -134,10 +135,13 @@ def get_table_df(
         self,
         table_name: str,
         schema_name: Optional[str] = None,
+        query_filter: Optional[str] = None,
```
natacha-scopeo, line 138: You need to filter in the get. 

The purpose of filtering by timestamp is to save time (less RAM consumption).

You must perform the filtering in the old and new df files, then synchronize the two concatenated df files (since the old identifiers no longer exist, they will not be deleted).

natacha-scopeo (approved):
It works locally 
Testing process : 
1. I ingest a snowflake table with a timestamp column, a text column and a metadata column
2. I update this snowflake table with a timestamp filtering (and same input attributes)

PR #75, opened 2025-07-16
[AIT-341] add gmail sender integration

natacha-scopeo:


ada_backend/services/graph/deploy_graph_service.py
```diff
@@ -56,7 +56,9 @@ def copy_component_instance(
             PipelineParameterSchema(name=parameter.name, value=parameter.value, order=parameter.order)
             for parameter in component_instance.parameters
         ],
+        integration=component_instance.integration,
     )
+    print("new_composant_instance", new_composant_instance)
```
jccscopeo, line 61: print to remove I think

ada_backend/services/graph/deploy_graph_service.py
```diff
@@ -79,6 +81,7 @@ def clone_graph_runner(
     old_relationships = []
     for component_node in graph_nodes:
         # Copy the component instance into a new component instance
+        print("component_node", component_node)
```
jccscopeo, line 84: print to remove

ada_backend/services/pipeline/update_pipeline_service.py
```diff
@@ -48,6 +52,15 @@ def create_or_update_component_instance(
     )
     instance_id = component_instance.id
 
+    if instance_data.integration and instance_data.integration.secret_id:
+        upsert_component_instance_integration(session, instance_id, instance_data.integration.secret_id)
+    elif instance_data.integration:
+        LOGGER.warning(
+            f"Integration provided for component instance {instance_id} but no secret ID found. "
+            "Deleting existing integration relationship."
```
jccscopeo, line 60: is it going to throw an error somewhere saying that the component is needing an integration in this case?

engine/integrations/gmail_sender.py
```diff
@@ -0,0 +1,146 @@
...
+            trace_manager,
+            tool_description=tool_description,
+            component_attributes=component_attributes,
+        )
+        if not google_client_id:
+            google_client_id = settings.GOOGLE_CLIENT_ID
+        if not google_client_secret:
+            google_client_secret = settings.GOOGLE_CLIENT_SECRET
+
+        session = next(get_db())
```
jccscopeo, line 84: We need to keep in my mind that this makes the agent not async compliant but not a problem at all with a few clients for performance I think

engine/integrations/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    return False
+
+
+def get_gmail_sender_service(access_token: str):
+    creds = Credentials(token=access_token)
+    service = build("gmail", "v1", credentials=creds)
+    return service
+
+
+def get_google_user_email(access_token: str) -> str:
```
jccscopeo, line 36: Can we make it async already?

engine/integrations/utils.py
```diff
@@ -0,0 +1,92 @@
...
+    resp = requests.get(url, headers=headers)
+
+    if resp.ok:
+        user_info = resp.json()
+        return user_info.get("email")
+    else:
+        raise ValueError(f"Failed to fetch user info: {resp.status_code} {resp.text}")
+
+
+def refresh_oauth_token(refresh_token: str, client_id: str, client_secret: str) -> tuple[str, str, datetime]:
```
jccscopeo, line 49: Maybe we can make it async as well

engine/integrations/gmail_sender.py
```diff
@@ -0,0 +1,146 @@
...
+        access_token = get_oauth_access_token(session, secret_integration_id, google_client_id, google_client_secret)
+        self.service = get_gmail_sender_service(access_token)
+        self.send_as_draft = send_as_draft
+
+        self.email_address = get_google_user_email(access_token)
+
+    def gmail_create_draft(self, email_subject: str, email_body: str, email_recipients: Optional[list[str]] = None):
```
jccscopeo, line 91: would be cool to see if we can make it async

jccscopeo (commented):
Left some comments. Could you give me a quick rundown on how the user stores his credentials and how we retrieve them for the components?

engine/integrations/gmail_sender.py
natacha-scopeo: ok it will be better to give the token as an input of the instance ?

engine/integrations/gmail_sender.py
```diff
@@ -0,0 +1,146 @@
...
+        try:
+            raw_email_message = create_raw_mail_message(
+                subject=email_subject,
+                body=email_body,
+                sender_email_address=self.email_address,
+                recipients=email_recipients,
+            )
+            draft = self.service.users().drafts().create(userId="me", body={"message": raw_email_message}).execute()
```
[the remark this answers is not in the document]
jccscopeo: no problem then

engine/integrations/gmail_sender.py
natacha-scopeo: I will not put access token in the graph description and I add as a todo to give the token as an input of the instance

jccscopeo (approved):
Working locally

PR #76, opened 2025-07-17
AIT-84-make gemini and mistral work for function call and RAG

dianascopeo:
Made Mistral and Gemini work when we use function call, rag, llm call.
Added OCR Call for parsing files

### DIFF
--- a/engine/agent/llm_call_agent.py
+++ b/engine/agent/llm_call_agent.py
@@ -9,6 +9,8 @@ from engine.agent.utils import extract_vars_in_text_template, parse_openai_messa
 from engine.llm_services.llm_service import CompletionService
 from engine.trace.trace_manager import TraceManager
 
+SUPPORTED_PROVIDER = "openai"
+
 
 class LLMCallAgent(Agent):
     def __init__(
@@ -79,7 +81,7 @@ class LLMCallAgent(Agent):
 
         if len(files_content) > 0:
             # TODO: Add support for other providers
-            if self._completion_service._provider != "openai":
+            if self._completion_service._provider != SUPPORTED_PROVIDER:
                 raise ValueError(f"File content is not supported for provider '{self._completion_service._provider}'.")
             content = [
                 {
--- a/engine/agent/llm_call_agent.py
+++ b/engine/agent/llm_call_agent.py
@@ -1,4 +1,3 @@
-import json
 from typing import Optional
 
 from openinference.semconv.trace import SpanAttributes
@@ -8,6 +7,7 @@ from engine.agent.agent import Agent, AgentPayload, ChatMessage, ComponentAttrib
 from engine.agent.utils import extract_vars_in_text_template, parse_openai_message_format
 from engine.llm_services.llm_service import CompletionService
 from engine.trace.trace_manager import TraceManager
+from engine.trace.serializer import serialize_to_json
 
 SUPPORTED_PROVIDER = "openai"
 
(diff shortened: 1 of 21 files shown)


engine/llm_services/llm_service.py
```diff
@@ -934,3 +991,50 @@ async def get_image_description_async(
...
+                    document={"type": "document_url", "document_url": f"data:application/pdf;base64,{pdf_content}"},
+                    include_image_base64=True,
+                )
+                full_document_markdown = ""
+                for page in ocr_response.pages:
+                    full_document_markdown += page.markdown
+                return full_document_markdown
+
+            case _:
+                raise ValueError(f"Invalid provider: {self._provider}")
```
marc-scopeo, line 1024: for OCR

engine/llm_services/llm_service.py
```diff
@@ -934,3 +991,50 @@ async def get_image_description_async(
...
+                client = mistralai.Mistral(api_key=self._api_key)
+                ocr_response = client.ocr.process(
+                    model="mistral-ocr-latest",
+                    document={"type": "document_url", "document_url": f"{pdf_content}"},
+                    include_image_base64=True,
+                )
+                full_document_markdown = ""
+                for page in ocr_response.pages:
+                    full_document_markdown += page.markdown
+                return full_document_markdown
```
marc-scopeo, line 1040: this is code duplication

also :line 1034 is different from previous version

engine/llm_services/utils.py
```diff
@@ -41,6 +41,38 @@ def chat_completion_to_response(
     return response_messages
 
 
+def clean_messages_for_mistral(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
```
marc-scopeo, line 44: don't say "clean". This is about making it compatible with mistral

engine/agent/ocr_call.py
```diff
@@ -0,0 +1,54 @@
...
+            ):
+
+                span = get_current_span()
+                span.set_attributes(
+                    {
+                        SpanAttributes.LLM_MODEL_NAME: self._ocr_service._model_name,
+                    }
+                )
+                response = await self._ocr_service.get_ocr_text_async(
+                    payload_json["messages"][-1]["content"][0]["file"]["file_data"]
```
natacha-scopeo, line 50: how can we be sure the payload is like this ? is this like to the ocr model we used ?

engine/llm_services/llm_service.py
```diff
@@ -604,6 +611,30 @@ def function_call(
                     stream=stream,
                     tool_choice=tool_choice,
                 )
+                print(response)
```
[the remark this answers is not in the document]
natacha-scopeo: still print

engine/llm_services/llm_service.py
```diff
@@ -697,6 +773,32 @@ async def function_call_async(
                     }
                 )
                 return response
+            case "mistral":
+                import openai
+
+                # Clean messages for Mistral's stricter API requirements
+                cleaned_messages = make_messages_compatible_for_mistral(messages)
```
natacha-scopeo, line 780: formatted_messages not cleaned (and you can remove the comment, function naming is enough)

engine/agent/llm_call_agent.py
```diff
@@ -78,6 +78,9 @@ async def _run_without_trace(self, *input_payloads: AgentPayload | dict) -> Agen
         text_content = self._prompt_template.format(**input_replacements)
 
         if len(files_content) > 0:
+            # TODO: Add support for other providers
+            if self._completion_service._provider != "openai":
```
natacha-scopeo, line 82: Can we create a global variable with the providers that support images ?

engine/llm_services/llm_service.py
```diff
@@ -697,6 +752,32 @@ async def function_call_async(
                     }
                 )
                 return response
+            case "mistral":
+                import openai
+
+                # Clean messages for Mistral's stricter API requirements
```
natacha-scopeo, line 758: remove comment

natacha-scopeo (approved):
it wroks locally ! not fan of the json output of the ocr

PR #78, opened 2025-07-17
[AIT-371] add component instance id

natacha-scopeo:


jccscopeo (approved):
It is working locally! I would like to have a change to have instead of "Agentic Reflexion" something like "{react_function_calling.name} is thinking"

dianascopeo (commented):
LGTM, only one question how we will handle old traces?

dianascopeo (approved):
Worked locally

PR #80, opened 2025-07-18
feat(Ingestion): Allow a user to update an already existing data source.

jccscopeo:
*feat(Ingestion):
- Allow the user to update an already existing data source
- The sync only update or append files, but does not suppress anything
- The task ingestion status is updated to **in_progress** when we enter the ingestion function
- The task ingestion status is  set to **updating_existing** if we are updating an existing source
- The ingestion task endpoint sends a 400 errors if the user attemps to create a new source with an already existing source name

### DIFF
--- a/ada_backend/routers/ingestion_task_router.py
+++ b/ada_backend/routers/ingestion_task_router.py
@@ -82,9 +82,10 @@ def update_organization_task(
         raise HTTPException(status_code=500, detail="Internal Server Error") from e
 
 
-@router.patch("/{organization_id}/update_by_source", status_code=status.HTTP_200_OK)
+@router.patch("/{organization_id}/sources/{source_id}/tasks", status_code=status.HTTP_200_OK)
 def update_ingestion_task_by_source(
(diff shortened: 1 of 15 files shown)


ada_backend/routers/ingestion_task_router.py
```diff
@@ -76,6 +81,37 @@ def update_organization_task(
         raise HTTPException(status_code=500, detail="Internal Server Error") from e
 
 
+@router.patch("/{organization_id}/update_by_source", status_code=status.HTTP_200_OK)
```
pabloscopeo, line 84: The route `PATCH /{organization_id}/update_by_source` could be more RESTful. Maybe we could do `PATCH /organizations/{org_id}/sources/{source_id}/tasks` instead? 

This way, it:
1. Follows the resource hierarchy (org -> source -> task)
2. Drops the action verb from the URL
3. Matches how tasks are actually linked to sources in the data model
4. Makes it clear which org's permissions we're checking

PR #82, opened 2025-07-18
Add some tracking of events

marc-scopeo:


ada_backend/segment_analytics.py
```diff
@@ -0,0 +1,104 @@
...
+    analytics.group(user_id=str(user_id), group_id=str(organization_id))
+
+
+@non_breaking_track
+def track_project_created(user_id: UUID, organization_id: UUID, project_id: UUID, project_name: str):
+    analytics.track(
+        user_id=str(user_id),
+        event="Project Created",
+        properties={
+            "machine_role": settings.MACHINE_ROLE,
```
natacha-scopeo, line 38: What is machine role ?

jccscopeo (commented):
Could we put a check before each tracking call if api_key is set up, so that local config customers do not have a lot of warning in their logs?

dianascopeo (approved):

PR #98, opened 2025-07-29
feat(NewTool): AIT-357-PDF generation tool

dianascopeo:
For Testing:

1. Run make db-seed
2. Add a PDF generation tool to your AI agent
3. At the end, the PDF file will be removed from the temp directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ custom_models.json
 
 # oauth
 ada-ingestion-system/.env
+
+# pdfs
+temp
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,3 @@ custom_models.json
 
 # oauth
 ada-ingestion-system/.env
-
-# temp files
-temp
--- a/.gitignore
+++ b/.gitignore
@@ -57,5 +57,5 @@ custom_models.json
 # oauth
 ada-ingestion-system/.env
 
-# pdfs
+# temp files
 temp
--- a/ada_backend/database/seed/seed_pdf_generation.py
+++ b/ada_backend/database/seed/seed_pdf_generation.py
@@ -1,0 +1,50 @@
+from uuid import UUID
+from sqlalchemy.orm import Session
+
+from ada_backend.database.models import (
+    Component,
+    ComponentParameterDefinition,
+    UIComponent,
+    UIComponentProperties,
+    ParameterType,
+)
+
+from ada_backend.database.component_definition_seeding import (
+    upsert_components,
+    upsert_components_parameter_definitions,
+)
+from ada_backend.database.seed.utils import COMPONENT_UUIDS
+from ada_backend.database.seed.seed_tool_description import TOOL_DESCRIPTION_UUIDS
+from ada_backend.database import models as db
+
+
+def seed_pdf_generation_components(session: Session):
+    pdf_generation_component = Component(
+        id=COMPONENT_UUIDS["pdf_generation"],
+        name="PDF Generation Tool",
+        description="Convert markdown text to PDF files.",
+        is_agent=False,
+        function_callable=True,
+        can_use_function_calling=False,
+        release_stage=db.ReleaseStage.BETA,
+        default_tool_description_id=TOOL_DESCRIPTION_UUIDS["default_pdf_generation_tool_description"],
+    )
+    upsert_components(session, [pdf_generation_component])
+
+    pdf_generation_parameter_definitions = [
+        ComponentParameterDefinition(
+            id=UUID("a1b2c3d4-e5f6-7890-abcd-ef1234567890"),
+            component_id=pdf_generation_component.id,
+            name="markdown_content",
+            type=ParameterType.STRING,
+            nullable=False,
+            ui_component=UIComponent.TEXTAREA,
+            ui_component_properties=UIComponentProperties(
+                label="Markdown Content",
+                placeholder="# Sample Markdown\n\nEnter your markdown content here...",
+                description="The markdown text to convert to PDF format.",
+            ).model_dump(exclude_unset=True, exclude_none=True),
+        ),
+    ]
+
+    upsert_components_parameter_definitions(session, pdf_generation_parameter_definitions)
--- a/ada_backend/database/seed/seed_pdf_generation.py
+++ b/ada_backend/database/seed/seed_pdf_generation.py
@@ -2,9 +2,10 @@ from sqlalchemy.orm import Session
 
 from ada_backend.database.models import Component
 
-from ada_backend.database.component_definition_seeding import upsert_components
+from ada_backend.database.component_definition_seeding import upsert_components, upsert_component_categories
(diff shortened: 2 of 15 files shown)


engine/graph_runner/graph_runner.py
```diff
@@ -159,6 +186,7 @@ async def _run_without_trace(self, *inputs: AgentPayload | dict, **kwargs) -> Ag
 
         # Collect outputs from leaf nodes
         final_output = self._collect_outputs()
+        delete_pdf_file_if_exists(final_output)
```
jccscopeo, line 189: Why do we call delete pdf function here?

engine/agent/pdf_generation_tool.py
```diff
@@ -0,0 +1,90 @@
...
+            success_msg = f"PDF generated successfully: {filename}"
+            LOGGER.info(success_msg)
+
+            return AgentPayload(
+                messages=[ChatMessage(role="assistant", content=success_msg)],
+                artifacts={"pdf_filename": str(filename)},
+                is_final=True,
+            )
+
+        except Exception as e:
```
jccscopeo, line 82: is it how we want to handle errors? Lookin at the other agents, we don t do try except in the agent structure I think

jccscopeo (commented):
Just some questions about how to handle pdfs errors and the mecanism of deletion of pdfs

engine/graph_runner/graph_runner.py
dianascopeo: we want at the end of process to remove pdf file

ada_backend/database/seed/seed_pdf_generation.py
```diff
@@ -0,0 +1,22 @@
...
+
+def seed_pdf_generation_components(session: Session):
+    pdf_generation_component = Component(
+        id=COMPONENT_UUIDS["pdf_generation"],
+        name="PDF Generation Tool",
+        description="Convert markdown text to PDF files.",
+        is_agent=False,
+        function_callable=True,
+        can_use_function_calling=False,
+        release_stage=db.ReleaseStage.BETA,
```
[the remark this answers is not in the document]
natacha-scopeo: it should be internal

ada_backend/database/seed/seed_pdf_generation.py
```diff
@@ -0,0 +1,22 @@
...
+        default_tool_description_id=TOOL_DESCRIPTION_UUIDS["default_pdf_generation_tool_description"],
+    )
+    upsert_components(session, [pdf_generation_component])
```
natacha-scopeo, line 22: Can you add category? action

engine/agent/pdf_generation_tool.py
```diff
@@ -0,0 +1,90 @@
...
+        if not markdown_content:
+            error_msg = "No markdown content provided"
+            LOGGER.error(error_msg)
+            return AgentPayload(
+                messages=[ChatMessage(role="assistant", content=error_msg)],
+                error=error_msg,
+                is_final=True,
+            )
+
+        uuid_suffix = str(uuid.uuid4())
```
natacha-scopeo, line 61: uuid suffix must be set in the graph level and add to the trace

natacha-scopeo: You need to create a folder with the uuid and then delete it in the finally of the try except

natacha-scopeo: I think it should be donne in the agent_runner file
```
#TODO: add the uuid
set_tracing_span(
        project_id=str(project_id),
        organization_id=str(project_details.organization_id),
        organization_llm_providers=get_organization_llm_providers(session, project_details.organization_id),
        conversation_id=input_data.get("conversation_id"),
    )
    try:
        agent_output = await agent.run(
            input_data,
            is_root_execution=True,
        )
    except Exception as e:
        raise ValueError(f"Error running agent: {str(e)}") from e
#TODO: add a finally to delete the temp folder
    return ChatResponse(
        message=agent_output.last_message.content, artifacts=agent_output.artifacts, error=agent_output.error
    )
```

ada_backend/services/agent_runner_service.py
```diff
@@ -133,6 +134,7 @@ async def run_agent(
             input_data,
             is_root_execution=True,
         )
+        delete_pdf_file_if_exists(agent_output)
     except Exception as e:
         raise ValueError(f"Error running agent: {str(e)}") from e
```
jccscopeo, line 139: Maybe we need to put a finally and delete_pdf_file_if_exist? I think we should create a function post_process agent run and we will update it to do post processes in the future (remove pdf here, but maybe other stuffs like remove temporary created files etc)

jccscopeo (commented):
I think we can create a post process function that we update when needed to add the clean up operations after agent run

.gitignore
```diff
@@ -56,3 +56,6 @@ custom_models.json
 
 # oauth
 ada-ingestion-system/.env
+
+# pdfs
```
natacha-scopeo, line 60: it is for all the temporary file not only pdf

ada_backend/services/agent_runner_service.py
natacha-scopeo: yes we need to a finally and delete there in case code fail

engine/agent/pdf_generation_tool.py
natacha-scopeo: why you cannot ?

dianascopeo: I think it is better to have in artifacts like tool created file and it put info file_path where is file and then we can get info for deletion

natacha-scopeo: temp folder must be at the level of one run of a graph. If we have 2 generation of pdfs, we will have 2 pdf in the same temp folder.

ada_backend/services/agent_runner_service.py
```diff
@@ -135,6 +136,7 @@ async def run_agent(
         )
     except Exception as e:
         raise ValueError(f"Error running agent: {str(e)}") from e
+    delete_pdf_file_if_exists(agent_output)
```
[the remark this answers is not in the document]
dianascopeo: I dont understand your comment

engine/agent/pdf_generation_tool.py
dianascopeo: in this case even we dont have pdf generation tool we will create folder everytime we run agent

ada_backend/services/agent_runner_service.py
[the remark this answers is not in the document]
natacha-scopeo: try except finally

engine/agent/pdf_generation_tool.py
natacha-scopeo: uuid yes but folder can only create if we need it and if it doesn't exist

natacha-scopeo (approved):
it works locally ! **Can we merge after release stage ?**

PR #104, opened 2025-07-31
feat(Code quality) : deduplicate qdrant code to avoid having same code twice on async and sync

jccscopeo:
- Pass every sync function as asyncio.run(async_version)
- Remove async test in testing as sync test are already running asyncio.run(async_version)
- at ingestion, a logging error happens at the end of the ingestion. It does not concern our script but the cleaning of httpx. Fixing this would involve refactoring all the ingestion code to make it async. Hence I keep it as it is now. It will only output an Event Close looped error log on the redis queue of the ingestion, but will not affect how the ingestion performs

### DIFF
--- a/tests/qdrant/test_qdrant_service.py
+++ b/tests/qdrant/test_qdrant_service.py
@@ -46,6 +46,7 @@ def test_qdrant_service():
     qdrant_agentic_service = QdrantService.from_defaults(
         embedding_service=embedding_service,
         default_collection_schema=qdrant_schema,
+        timeout=60.0,  # Increased timeout for tests
     )
 
     # Ensure a clean state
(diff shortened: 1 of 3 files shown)


tests/qdrant/test_qdrant_service.py
```diff
@@ -350,99 +229,3 @@ def test_qdrant_filtering(
     assert expected_chunk == set_chunks
     assert qdrant_agentic_service.delete_collection(TEST_COLLECTION_NAME)
     assert not qdrant_agentic_service.collection_exists(TEST_COLLECTION_NAME)
-
-
-@pytest.mark.asyncio
```
natacha-scopeo, line 355: Why removing test ?

jccscopeo: I remove async tests and kept sync tests because now every sync function is an asyncio.run(async_test) so we did not need to retest again the async function via async test. And also the main reason is that testing the async tests were throwing networks errors because it tries to have multiple connections at the same time and our free qdrant plan does not allow us to have this much use at the same time

pabloscopeo (approved):
I don't think we need to keep the sync methods. They just to `asyncio.run(...)` which can be done outside. I think that's ok for 1 main method for it's probably not necessary for this many.

Also, at the end did we come up with a convention on the async naming yet? I think it's better if the methods name's are simply something like: `async def get_collection_data`, there's probably no need to suffix with "async" again.

Code wise LGTM. Still, it would be nice if we could harmonize this before merging.

PR #106, opened 2025-07-31
AIT-436-handle hour in penalty function

dianascopeo:


engine/qdrant_service.py
```diff
@@ -377,7 +377,8 @@ def apply_date_penalty_to_chunks(
             if not date:
                 penalized_score = default_penalty_rate
             else:
-                chunk_date = datetime.strptime(date, "%Y-%m-%d")
+                # TODO: Handle all cases and make general function
+                chunk_date = datetime.strptime(date, "%Y-%m-%d %H:%M:%S")
```
natacha-scopeo, line 381: Why changing ?

dianascopeo: because date format chnaged in cogiterra

dianascopeo: now we have hours as well

natacha-scopeo: and for other column ?

PR #108, opened 2025-07-31
Ait 439 renaming of functions and file and small bugfixes

marc-scopeo:


ada_backend/repositories/component_repository.py
```diff
@@ -265,90 +265,92 @@ def get_all_components_with_parameters(
...
+                            id=param_child_def.child_component_id,
+                            parameter_name=subcomponent_param.name,
+                            is_optional=subcomponent_param.nullable,
+                        )
+                        for subcomponent_param, param_child_def in subcomponent_params
+                    ],
+                    categories=get_categories(session, component.id),
+                )
+            )
+        except Exception as e:
```
pabloscopeo, line 352: Can we add a TODO message saying that we should put this somewhere else?

pabloscopeo (approved):
Code-wise, LGTM

ada_backend/repositories/component_repository.py
marc-scopeo: I think we want to catch everything here. If we don't we break all the components

natacha-scopeo: Is it possible that a break in the a component will send us an alert ? but the get_component function will still work ?

natacha-scopeo (approved):
It works locally! I test to run RAG, LLM Call and AI agent with internet search and api call.

ada_backend/repositories/component_repository.py
marc-scopeo: We have an error log... we could make alerts based on error level logs

PR #109, opened 2025-08-01
Created batching directives and used them in GR

pabloscopeo:


ada_backend/database/models.py
```diff
@@ -118,6 +118,15 @@ class ReleaseStage(StrEnum):
     INTERNAL = "internal"
 
 
+class ComponentType(StrEnum):
+    """Enumeration of component types for classification."""
+
+    GRAPH_RUNNER = "graph_runner"
+    AGENT = "agent"                        # Standard computational agents
```
marc-scopeo, line 125: In another PR of the summer, I'd like to cal that something else and keep the word agent for the meaning that the AI community is giving.

ada_backend/services/registry.py
```diff
@@ -418,6 +423,21 @@ def create_factory_registry() -> FactoryRegistry:
         ),
     )
 
+    # Control Directives - use EntityFactory since they don't need trace manager
+    registry.register(
+        name=SupportedEntityType.BATCH_START,
```
marc-scopeo, line 428: For the novice, this is not understandable. "Chunking" is most understandable term, and batch processing should be implicit in the UI/UX.

marc-scopeo (commented):
There are very little differences with my previous work, except that you made the chunker and splitter something else than ordinary components (I don't want to use the word agent anymore)

PR #112, opened 2025-08-05
feat(Local Version):Add new supabase migration and functions for local set up of supabase

jccscopeo:
**PR** that aims at putting the new superadmin console of supabase in the local version + add template organization on the local version.

**How to test**:
- Set up the local version with supabase (follow the readme)
- check with the front that you have the superadmin console visible on the front app
- check that you can create a project, deploy it in production and use it as a template


dianascopeo (commented):
is there a way to test this?

PR #117, opened 2025-08-08
AIT-359-add env and call type into trace attributes and trace endpoint

dianascopeo:


ada_backend/routers/project_router.py
```diff
@@ -201,11 +203,20 @@ async def get_project_trace(
     project_id: UUID,
     duration: int,
     user: Annotated[SupabaseUser, Depends(get_user_from_supabase_token)],
+    environment: EnvType = None,
+    call_type: CallType = None,
```
pabloscopeo, line 207:
```suggestion
    environment: Optional[EnvType] = None,
    call_type: Optional[CallType] = None,
```

ada_backend/routers/project_router.py
```diff
@@ -219,11 +230,20 @@ async def get_project_trace_v2(
...
```
pabloscopeo, line 234:
```suggestion
    environment: Optional[EnvType] = None,
    call_type: Optional[CallType] = None,
```

ada_backend/services/trace_service.py
```diff
@@ -159,7 +162,12 @@ def build_span_trees(df: pd.DataFrame, include_messages: bool) -> List[TraceSpan
 
 
 def get_trace_by_project(
-    user_id: UUID, project_id: UUID, duration: int, include_messages: bool = False
+    user_id: UUID,
+    project_id: UUID,
+    duration: int,
+    include_messages: bool = False,
+    environment: EnvType = None,
+    call_type: CallType = None,
```
pabloscopeo, line 170:
```suggestion
    environment: Optional[EnvType] = None,
    call_type: Optional[CallType] = None,
```

engine/trace/sql_exporter.py
```diff
@@ -149,6 +149,23 @@ def export(self, spans: list[ReadableSpan]) -> SpanExportResult:
...
+            if not environment:
+                environment = span.attributes.get("environment")
+            if not call_type:
+                call_type = span.attributes.get("call_type")
+
+            # Remove them from attributes since they're now stored in dedicated columns
+            if "environment" in formatted_attributes:
+                del formatted_attributes["environment"]
+            if "call_type" in formatted_attributes:
+                del formatted_attributes["call_type"]
```
pabloscopeo, line 167: I'm not sure to understand, so the values can come from attributes and columns? I thought that maybe we could have them only in the columns to simplify. Or is this like this because the only way we can carry the new values is by putting them on the pandas df that we processed and then we get everything as `formatted_attributes`? 

In any case, as I understand, in this PR we are only storing new data in columns right? I think I'm just a bit confused on the entire flow on how this is done.

dianascopeo: yes you are right I will remove if else part

pabloscopeo (approved):
Works locally! Tested interacting via API and Sandbox, and then retrieved traces filtering by `call_type`.

PR #118, opened 2025-08-17
AIT-451 Project Reference Block

pabloscopeo:
# PR: Secure Project Referencing (ProjectReference / GraphRunnerBlock)

## Why

This PR lays the groundwork for the upcoming **Chunk Processor**. The main challenges were not in splitting/merging data, but in **secure project referencing**: how to run another project’s graph as a component, ensure the user has permission, and handle sync/async safely. Once this is solid, the Chunk Processor can reuse the same parameter processor and focus only on chunking logic.

## What’s in this PR

* **GraphRunnerBlock** agent to run a graph as a single node (supports nested graphs).
* **ProjectReference** component (UI-facing) with `project_id`.
* **RequestContext** via `ContextVar` for Supabase user per request.
* **Permission validation** in parameter processor using Supabase + `anyio.from_thread.run`.
* **Entity factory/registry wiring** for ProjectReference → GraphRunnerBlock.
* **Seeding**: component, tool description, constants; registry entries.
* **Fixes**: circular imports (lazy imports, constants refactor) (**edits on seed files**).
* **Tests**: GraphRunnerBlock incl. nested graphs & IO handling.

## Tested Flow

Built a flow on the frontend:
`API Input → AI Agent (internet search) → ProjectReference (child project: simple LLM call that translates to sarcastic French`
Input: “tell me the current conversion rate of the dollar.”
Result: worked end-to-end — internet search produced rates, ProjectReference invoked the child project, and the final output was the sarcastic French translation.
**NOTE: Referenced project needs to be _deployed_ to make it work, ProjectReference uses the latest _production_  project referenced. Choosing the version will come soon**).

## Debug note

When running a ProjectReference, the output from the upstream block was being lost upon entering the child project’s **Input** component, which defaulted to `"Hello"`. This happened because the Input logic overwrote incoming values with defaults and only worked with `dict`, while the ProjectReference passed an `AgentPayload`. The fix was to update the Input block to handle both `AgentPayload` and `dict`, preserving runtime values instead of replacing them. With this change, inputs now flow correctly into child graphs, and ProjectReference produces the expected output.

### DIFF
--- a/ada_backend/services/entity_factory.py
+++ b/ada_backend/services/entity_factory.py
@@ -1,5 +1,8 @@
 import logging
 import json
+import asyncio
+import concurrent.futures
+import contextvars
 from inspect import signature
 from typing import Optional, Any, Type, Callable, get_type_hints, get_origin, get_args, Union
 from uuid import UUID
@@ -11,7 +14,11 @@ from engine.trace.trace_context import get_trace_manager
 from engine.llm_services.llm_service import EmbeddingService, CompletionService, WebSearchService, OCRService
 from engine.qdrant_service import QdrantService, QdrantCollectionSchema
 from ada_backend.database.setup_db import get_db_session
+from ada_backend.database.models import EnvType
 from ada_backend.repositories.source_repository import get_data_source_by_id
+from ada_backend.repositories.project_repository import get_project
+from ada_backend.context import get_request_context
+from ada_backend.services.user_roles_service import get_user_access_to_organization
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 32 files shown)


ada_backend/services/entity_factory.py
```diff
@@ -405,6 +412,92 @@ def processor(params: dict, constructor_params: dict[str, Any]) -> dict:
...
+        env = EnvType(env_str)
+
+        context = get_request_context()
+        user = context.require_user()
+
+        # Capture the current context including TraceManager.
+        current_context = contextvars.copy_context()
+
+        with get_db_session() as session:
+            # TODO: Fix circular import issue - these imports are here to avoid
```
marc-scopeo, line 446: very important indeed

engine/agent/graph_runner_block.py
```diff
@@ -0,0 +1,39 @@
...
+        self._graph_runner = graph_runner
+
+    async def _run_without_io_trace(self, *inputs: AgentPayload, **kwargs) -> AgentPayload:
+        input_data: AgentPayload = inputs[0]
+
+        result = await self._graph_runner.run(input_data)
+
+        # TODO: This should be enforced on the GraphRunner level
+        if not isinstance(result, AgentPayload):
+            raise ValueError("GraphRunnerBlock must return an AgentPayload")
```
marc-scopeo, line 37: why this check ? We are not going to check all types manually

natacha-scopeo (commented):
it is working locally!

PR #122, opened 2025-08-20
[AIT-429] create 2 endpoints to get traces

natacha-scopeo:
Creation of 2 endpoints for observability
- to **retrieve root spans**: `projects/{project_id}/traces`
  output format a list of `RootSpan`
  ```
  class RootTraceSpan(BaseModel):
      trace_id: str
      span_id: str
      name: str
      span_kind: str
      start_time: str
      end_time: str
      input: list
      output: list
      status_code: str
      cumulative_llm_token_count_prompt: int
      cumulative_llm_token_count_completion: int
      llm_token_count_prompt: int | None
      llm_token_count_completion: int | None
  ```
- an endpoint to **retrieve information from a specific trace**
 `traces/{trace_id}/tree` => same output format as before

### DIFF
--- a/ada_backend/services/trace_service.py
+++ b/ada_backend/services/trace_service.py
@@ -1,6 +1,6 @@
 from collections import defaultdict
 import json
-from typing import List
+from typing import List, Optional
 from uuid import UUID
 import logging
 
@@ -243,8 +243,21 @@ def get_span_trace_service(user_id: UUID, trace_id: UUID) -> TraceSpan:
     return span_trees[0]
 
 
-def get_root_traces_by_project(user_id: UUID, project_id: UUID, duration: int) -> List[RootTraceSpan]:
+def get_root_traces_by_project(
+    user_id: UUID,
+    project_id: UUID,
+    duration: int,
+    environment: Optional[EnvType] = None,
+    call_type: Optional[CallType] = None,
+) -> List[RootTraceSpan]:
     df_span = query_root_trace_duration(project_id, duration)
     track_project_observability_loaded(user_id, project_id)
     LOGGER.info(f"Querying root spans for project {project_id} with duration {duration} days")
+
+    if environment is not None:
+        df_span = df_span[df_span["environment"] == environment.value]
+
+    if call_type is not None:
+        df_span = df_span[df_span["call_type"] == call_type.value]
+
     return build_root_spans(df_span)
--- a/ada_backend/routers/trace_router.py
+++ b/ada_backend/routers/trace_router.py
@@ -1,9 +1,10 @@
 import logging
-from typing import Annotated, List
+from typing import Annotated, List, Optional
 from uuid import UUID
 
 from fastapi import APIRouter, Depends, HTTPException
 
+from ada_backend.database.models import CallType, EnvType
 from ada_backend.routers.auth_router import get_user_from_supabase_token
 from ada_backend.schemas.auth_schema import SupabaseUser
 from ada_backend.schemas.trace_schema import RootTraceSpan, TraceSpan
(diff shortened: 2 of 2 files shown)


ada_backend/services/metrics/utils.py
```diff
@@ -46,3 +46,37 @@ def query_trace_messages(
...
+    session = get_session_trace()
+    df = pd.read_sql_query(query, session.bind)
+    session.close()
+    df = df.replace({np.nan: None})
+    df["attributes"] = df["attributes"].apply(lambda x: json.loads(x) if isinstance(x, str) else x)
+    df_expanded = df.join(pd.json_normalize(df["attributes"]))
+    trace_rowids = df_expanded[df_expanded["parent_id"].isna() & (df_expanded["project_id"] == str(project_id))][
+        "trace_rowid"
+    ].values
+    return df[(df["trace_rowid"].isin(trace_rowids)) & (df["parent_id"].isna())]
```
pabloscopeo, line 68: maybe worth considering doing this fully in SQL with a parameterized query, ex: filter by project_id, parent_id IS NULL, and the time window directly in the DB, and use a CTE with MAX(start_time) for ordering. that should be more efficient if this endpoint gets hit often. pandas is nice for debugging though, so I get the trade-off. if it feels like too much change right now, we could just leave a small TODO for later.

ada_backend/services/metrics/utils.py
```diff
@@ -46,3 +46,37 @@ def query_trace_messages(
...
+    query = (
+        "SELECT s.*, m.input_content,m.output_content FROM spans s "
+        f"LEFT JOIN span_messages m ON m.span_id = s.span_id WHERE s.trace_rowid = '{trace_id}';"
+    )
+    session = get_session_trace()
+    df = pd.read_sql_query(query, session.bind)
+    session.close()
+    df = df.replace({np.nan: None})
+    print(df.columns)
+    print(df.head())
```
pabloscopeo, line 81: prints

ada_backend/services/metrics/utils.py
pabloscopeo, line 1: is this really a “utils”? these functions feel more like DB access (trace repository). maybe they could take a session as param, skip pandas, and return tailored DTOs, would make things more structured/predictable. that said, pandas is handy and I’m fine if we want to accept the tech debt here for now.

ada_backend/services/trace_service.py
```diff
@@ -179,3 +215,17 @@ def get_token_usage(organization_id: UUID) -> TokenUsage:
     if not token_usage:
         return TokenUsage(organization_id=str(organization_id), total_tokens=0)
     return TokenUsage(organization_id=token_usage.organization_id, total_tokens=token_usage.total_tokens)
+
+
+def get_span_trace_service(user_id: UUID, trace_id: UUID) -> TraceSpan:
+    df_span = query_trace_by_trace_id(trace_id)
+    track_span_observability_loaded(user_id, trace_id)
+
+    return build_span_trees(df_span, include_messages=True)[0]
```
pabloscopeo, line 224: could this return an empty list?

ada_backend/services/metrics/utils.py
natacha-scopeo: I agree for a TODO but as it will change the db I think we need to do it in a specific PR

ada_backend/services/trace_service.py
natacha-scopeo: It shouldn't be possible but I will add a check

pabloscopeo (approved):
Tested and works locally ✅
Before merging, I suggest we:
1. Mark old endpoints with `deprecated=True`.
2. Add TODOs:
   * Move the "utils" functions into the repository layer (note in the module docstring).
   * Refactor these functions to be SQL-native for better efficiency.

ada_backend/segment_analytics.py
```diff
@@ -109,6 +109,18 @@ def track_project_observability_loaded(user_id: UUID, project_id: UUID):
     )
 
 
+@non_breaking_track
+def track_span_observability_loaded(user_id: UUID, span_id: UUID):
+    analytics.track(
+        user_id=str(user_id),
+        event="Observability Loaded",
+        properties={
+            "env": settings.ENV,
```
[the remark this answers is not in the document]
natacha-scopeo: draft or prod

dianascopeo (approved):
Worked locally using docs

pabloscopeo (approved):

PR #124, opened 2025-08-22
[AIT-452] add gpt-5 to models

natacha-scopeo:


marc-scopeo (approved):

ada_backend/database/seed/seed_ai_agent.py
```diff
@@ -18,7 +18,7 @@
     ParameterLLMConfig,
     build_function_calling_service_config_definitions,
 )
-from ada_backend.services.registry import COMPLETION_MODEL_IN_DB
+from ada_backend.services.registry import COMPLETION_MODEL_IN_DB, TEMPERATURE_IN_DB
```
pabloscopeo, line 21: This will have circular imports later. I think it's better if we do seed -> db (models) -> repo -> service (registry lives here). These import do service -> seed making an import cycle. I think it's better if we do it like this: https://github.com/Scopeo/draftnrun/pull/118/files#diff-577f358bd15977dc06523cff60df7cc0d2732fe5be7c93591bc5028fc96f4a8a

ada_backend/database/seed/utils.py
```diff
@@ -146,15 +146,66 @@ def build_completion_service_config_definitions(
...
                     default="1.0",
+                    ui_component=UIComponent.TEXTFIELD,
+                    ui_component_properties=UIComponentProperties(
+                        label="Temperature",
+                        placeholder="Enter temperature, it is different for each model, check the model documentation",
+                    ).model_dump(exclude_unset=True, exclude_none=True),
+                    is_advanced=True,
+                )
+            )
+        if param.param_name == "verbosity":
```
pabloscopeo, line 166: Is it ok if this is hardcoded? what are the possible values that can come from `param.param_name` and where are them defined?

pabloscopeo (commented):
Small issue to check about circular imports, but it's important imo

ada_backend/database/seed/utils.py
dianascopeo: done

ada_backend/database/seed/seed_ai_agent.py
pabloscopeo: yes that will solve this issue and help with conflicts later too

natacha-scopeo (commented):
LGMT

PR #132, opened 2025-09-03
[dra-371] send pdf by email

natacha-scopeo:
# Add CC/BCC and attachment support to GmailSender

- Enables adding recipients in copy (Cc) and blind copy (Bcc).
- Adds an attachments parameter to include one or more files.
- Propagates the new parameters through `create_raw_mail_message`, `gmail_send_email`, `gmail_create_draft`, and `_run_without_io_trace`.


dianascopeo (approved):
Worked locally

PR #134, opened 2025-09-04
Change qdrant filtering for rag

dianascopeo:


dianascopeo:
> We need to be careful as now we don t have any tests on the filtering. Maybe we should keep the tests that are existing just to be sure that passing the correct filter will make it work

I can create new function I removed old one as now we dont have function that creates filter

natacha-scopeo (approved):
it is working locally!

PR #135, opened 2025-09-04
feat(dbingestionfix): fix db ingetsion not create a new source when we update

dianascopeo:


ingestion_script/utils.py
```diff
@@ -86,8 +86,9 @@ def create_source(
             },
         )
         response.raise_for_status()
-        LOGGER.info(f"Successfully created source for organization {organization_id}")
-        return response.json()
+        source_id = response.json()
+        LOGGER.info(f"Successfully created source {source_id} for organization {organization_id}")
+        return str(source_id)
```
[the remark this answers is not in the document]
dianascopeo: I will rollback now this part

ingestion_script/utils.py
```diff
@@ -73,7 +73,7 @@ def update_ingestion_task(
 def create_source(
     organization_id: str,
     source_data: DataSourceSchema,
-) -> None:
+) -> str:
```
natacha-scopeo, line 76: still str ? not uuid ?

ingestion_script/utils.py
```diff
@@ -96,6 +97,47 @@ def create_source(
         ) from e
 
 
+def upsert_source(
+    organization_id: str,
+    source_data: DataSourceSchema,
+    source_id: Optional[str] = None,
+) -> str:
```
natacha-scopeo, line 104: idem

natacha-scopeo (commented):
It is working locally with docs

ada_backend/repositories/ingestion_task_repository.py
```diff
@@ -32,13 +32,15 @@ def create_ingestion_task(
     source_name: str,
     source_type: db.SourceType,
     status: db.TaskStatus,
+    source_id: Optional[UUID] = None,
```
MelMayssonOwen, line 35: why optional if it's only for the update ?

update and create are two different routes.

create needs no id and update needs one

@natacha-scopeo

natacha-scopeo: This is in create ingestion task (same ingestion task/ same endpoint for create and update)

jccscopeo (approved):
LGTM, Working locally

ingestion_script/utils.py
```diff
@@ -96,6 +98,27 @@ def create_source(
...
 
+def upsert_source(
+    organization_id: str,
+    source_data: DataSourceSchema,
+    source_id: Optional[UUID] = None,
+) -> UUID:
+    """Create or update a source in the database."""
+
+    try:
+        if source_id:
```
natacha-scopeo, line 109: can you put this if else outside this function (and delete this function). 
We only create source. Wa cannot modify a source

natacha-scopeo (approved):
it is working locally

PR #153, opened 2025-09-19
[dra-433] Bugfix: attributes in obs

natacha-scopeo:


jccscopeo (approved):
Working locally for all trace endpoints

PR #155, opened 2025-09-19
Feat(GraphRunnerTags): Add tag to graph runner

dianascopeo:
### Summary
This PR introduces a versioning/tagging system for GraphRunner objects in the backend. The new feature allows every deployed graph runner to be associated with a semantic version tag (e.g., v0.0.1), supports tag-based chat endpoint.

### Key Changes
Database & Models

- GraphRunner Model:
Added a nullable tag_version column to GraphRunner (SQLAlchemy & Alembic migration).

- Spans Table:
Added a nullable tag_version column to the Span model (SQLAlchemy & Alembic migration).

### How to test (using docs)
Update your tables
`make db-upgrade`
`make trace-db-upgrade`
Test 2 endpoints:
`GET projects/{project_id}` to get all graph runners info
`POST projects/{project_id}/graphs/{graph_runner_id}/chat`


ada_backend/routers/tag_router.py
```diff
@@ -0,0 +1,33 @@
...
+        SupabaseUser, Depends(user_has_access_to_project_dependency(allowed_roles=UserRights.READER.value))
+    ],
+    session: Session = Depends(get_db),
+):
+    if not user.id:
+        raise HTTPException(status_code=400, detail="User ID not found")
+    try:
+        return list_tag_versions_service(session, project_id)
+    except Exception as e:
+        raise HTTPException(status_code=500, detail="Internal Server Error") from e
```
pabloscopeo, line 33: The resource is the project; tags/versions are a sub-resource. This should be `GET /projects/{id}/tags` (or a tags field on `GET /projects/{id}`), not a top-level “tags” endpoint (currently it's `/tags/{project_id}`). 
Filtering projects by tag would be `GET /projects?tag=….` (for example, if we ever needed it).

ada_backend/services/agent_runner_service.py
```diff
@@ -112,13 +114,35 @@ async def run_env_agent(
...
 async def run_agent(
     session: Session,
     project_id: UUID,
     graph_runner_id: UUID,
     input_data: dict,
-    environment: EnvType,
-    call_type: CallType,
+    environment: Optional[EnvType] = None,
+    call_type: Optional[CallType] = None,
+    tag_version: Optional[str] = None,
```
pabloscopeo, line 145: - Shouldn't `call_type` say mandatory?
- Currently there are many different ways to use this service function:
  - directly using the GR's id
  - using env
  - using tag version

Maybe we could add a check to enforce that only one of them can be provided, to avoid conflicts.

ada_backend/services/graph/get_graph_service.py
```diff
@@ -66,8 +67,14 @@ def get_graph_service(
         )
         LOGGER.info(f"Edge from {edge.source_node_id} to {edge.target_node_id}")
 
+    try:
+        tag_version = get_graph_runner_tag_version(session, graph_runner_id)
+    except Exception:
+        tag_version = None
```
pabloscopeo, line 73: Why this?

ada_backend/services/tag_service.py
```diff
@@ -0,0 +1,8 @@
+from uuid import UUID
+from sqlalchemy.orm import Session
+
+from ada_backend.repositories.tag_repository import list_tag_versions
+
+
+def list_tag_versions_service(session: Session, project_id: UUID) -> list[str]:
+    return list_tag_versions(session, project_id)
```
pabloscopeo, line 8: This is perfect imo. But I don't think we need a tag_service file. The real entity are projects. Tags are just attributes imo.

ada_backend/routers/project_router.py
```diff
@@ -297,14 +297,15 @@ async def chat(
         env_relationship = get_env_relationship_by_graph_runner_id(session, graph_runner_id)
```
natacha-scopeo, line 297: In this function can we not get tag and env at the same time ?

ada_backend/services/graph/get_graph_service.py
```diff
@@ -66,8 +67,11 @@ def get_graph_service(
         )
         LOGGER.info(f"Edge from {edge.source_node_id} to {edge.target_node_id}")
 
+    tag_version = get_graph_runner_tag_version(session, graph_runner_id)
```
natacha-scopeo, line 70: same here can we not concatenate with `get_env_relationship_by_graph_runner_id` function

ada_backend/routers/project_router.py
natacha-scopeo: maybe rename variable and function ?

natacha-scopeo (commented):
I think it will be nice to add a tag to all the graph that already exist ? starting form v0.01 for the older to the most recent

natacha-scopeo (approved):

PR #160, opened 2025-09-24
feat(TraceDB): move project_id to new column and change some fucntions to use directly sql query

dianascopeo:
### Overview

This PR migrates project_id from being stored in JSON attributes to a dedicated indexed column in the spans table for improved query performance and data structure.

### How to test
`make trace-db-upgrade`
Test different projects and check observability


natacha-scopeo (commented):
Code-wise, it looks good!
I will test

jccscopeo (approved):
Working locally with AI agent + LLM Call, RAG and Internet search tools.
Downgrading + running on main the trace still works, ugrading + running this branch also

natacha-scopeo (approved):
no duplicated project_id

PR #163, opened 2025-09-24
Pablo/dra 441 agent io refactor mvp

pabloscopeo:
### **feat: Multi-Port I/O Refactor for Graph Engine**

This PR introduces a fundamental refactor to the graph execution engine, moving from a single-stream I/O model to a flexible multi-port system. This allows any component to be a first-class graph node and enables complex, declarative workflows.

#### **Core Concepts & Changes**

*   **`PortDefinition` & `PortMapping`**: We've introduced two new DB tables. `port_definitions` stores the I/O schema for each component type (e.g., "this component has a 'query' input and a 'response' output"). `port_mappings` stores the instance-level "wiring" for a specific graph, connecting an output port of one node to an input port of another.

*   **Code-First I/O Schemas**: Component I/O is now defined directly in the code using Pydantic models via `get_inputs_schema()` and `get_outputs_schema()` methods on the `Agent` class. A new seeding script introspects these methods and automatically populates the `port_definitions` table, ensuring the DB is always in sync with the code.

*   **`NodeData` Packet**: A new universal data packet, `NodeData`, now flows between nodes. It contains a `data` dictionary for port values and a shared `ctx` dictionary for runtime context. This is intended to eventually replace the legacy `AgentPayload`.

*   **`GraphRunner` Refactor**: The `GraphRunner` is now driven by `port_mappings`. The `_gather_inputs` method assembles a node's inputs based on these explicit connections, providing a clean separation between execution dependency (edges) and data flow (mappings).

#### **Type Coercion System**

A `CoercionMatrix` automatically handles type mismatches between components at runtime. It uses a simple priority system: pass-through for correct types, direct coercers for specific pairs (e.g., `ChatMessage` → `str`), primitive fallbacks, and clear errors for impossible conversions. This ensures components can be connected regardless of their I/O types while maintaining type safety.

**Key coercers:**
- `list[ChatMessage] → str` (extracts last message content)
- `AgentPayload → str` (legacy compatibility)
- `dict → str` (extracts common fields like content, output)

The system validates port mappings at build-time with strict validation for migrated components and lenient warnings for legacy ones, supporting gradual migration.

#### **API & Backward Compatibility**

*   **API Update**: The `PUT /projects/{id}/graph/{id}` endpoint's request body now accepts a `port_mappings` array, allowing the frontend to save the graph's wiring.

*   **Phased Migration**: A `migrated: bool` flag has been added to the base `Agent` class.
    *   If `True`, the component uses the new typed `_run_without_io_trace(inputs: PydanticModel, ...)` method.
    *   If `False`, it uses the legacy path. The `Agent.run()` method acts as a dispatcher.

*   **Backward Compatibility**: All compatibility logic is isolated within the `GraphRunner` and base `Agent` class. The runner normalizes outputs from legacy components (which may return `dict` or `AgentPayload`) into the standard `NodeData` packet, ensuring the core loop remains clean. `AgentPayload` and its related shims can be safely removed once all components are migrated.

#### **UX & Future Features**

*   **Canonical Ports & Auto-Wiring**: 
    - **Runtime Resolution (Fallback)**: The GraphRunner now automatically falls back to canonical ports when an edge lacks explicit mappings. If no mapping is provided, it connects the "canonical output" of the source node to the "canonical input" of the target node. This ensures legacy graphs run without modification.

    - **Save-Time Resolution (Persistence)**: The update_graph_service now also persists these defaults. When saving a graph, if an edge has no explicit mapping, the service resolves it using each component’s get_canonical_ports() and writes the mapping to the DB. This guarantees that when the graph is reloaded, all mappings are explicit and unambiguous.

    - **NOTE:** This also helps retro-compatibility and migration:
      -> An old project without port mappings will still run thanks to runtime resolution.
      -> If that old project is simply saved, the backend will automatically resolve its port mappings and transparently migrate it.

*   **Dispatch Strategies**: `PortMapping` includes a `dispatch_strategy` field. The `direct` strategy (a simple value copy) is fully implemented. A placeholder for a `function_call` strategy has been added to the `GraphRunner`, making it clear where to implement future LLM-based data structuring.

**IMPORTANT: To test with frontend, checkout `pablo/dra-441-agent-io-refactor-mvp` there too**

### DIFF
--- a/ada_backend/database/seed/seed_filter.py
+++ b/ada_backend/database/seed/seed_filter.py
@@ -105,8 +105,7 @@ def seed_filter_components(session: Session):
             ),
         ],
     )
-    # TODO: Delete manual port seeding after Filter component is migrated
-    # Upsert canonical port definitions for Filter component
+    # LEGACY: Manual port seeding for unmigrated Filter component
     existing = session.query(db.Component).filter(db.Component.id == filter.id).first()
     if existing:
         # Ensure INPUT and OUTPUT canonical 'messages' ports exist
--- a/ada_backend/database/seed/seed_input.py
+++ b/ada_backend/database/seed/seed_input.py
@@ -38,7 +38,7 @@ def seed_input_components(session: Session):
             input,
         ],
     )
-    # Upsert canonical port definitions for Input component
+    # LEGACY: Manual port seeding for unmigrated Input component
     existing = get_component_by_id(session, input.id)
     if existing:
         # Ensure an OUTPUT canonical 'messages' port exists
--- a/engine/agent/agent.py
+++ b/engine/agent/agent.py
@@ -16,6 +16,7 @@ from engine.agent.types import (
     AgentPayload,
     ChatMessage,
 )
+from engine import legacy_compatibility
 from engine.trace.trace_manager import TraceManager
 from engine.trace.serializer import serialize_to_json
 from engine.prometheus_metric import track_calls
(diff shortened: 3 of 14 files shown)


ada_backend/database/seed/integrations/seed_gmail.py
```diff
@@ -23,7 +23,7 @@ def seed_gmail_components(session: Session):
         id=COMPONENT_UUIDS["gmail_sender"],
         name="Gmail Sender",
         description="A component to send emails using Gmail API.",
-        is_agent=False,
+        is_agent=True,
```
marc-scopeo, line 26: shouldn't we remove this if everybody is now an "agent" (a node, to use the new term) ?

marc-scopeo (commented):
Without getting in the details, I like it.

ada_backend/database/seed_db.py
```diff
@@ -34,12 +34,91 @@
...
 from ada_backend.database.seed.seed_tool_description import seed_tool_description
 from ada_backend.database.seed.utils import COMPONENT_UUIDS
+from ada_backend.services.registry import FACTORY_REGISTRY
+from engine.agent.agent import Agent
+
 
 LOGGER = logging.getLogger(__name__)
 
 
+def seed_port_definitions(session: Session):
```
natacha-scopeo, line 47: can put this function in a dedicated file ?

ada_backend/services/graph/update_graph_service.py
```diff
@@ -126,3 +164,119 @@ async def update_graph_service(
...
+    canonical_ports_by_component = get_canonical_ports_for_components(session, component_ids)
+
+    auto_mappings: list[db.PortMapping] = []
+    for source_iid, target_iid in unmapped_edges:
+        if source_iid not in instance_to_component or target_iid not in instance_to_component:
+            raise ValueError(
+                "Unable to infer component ids for one or more unmapped edges; please provide explicit port_mappings."
+            )
+
+        source_cid = instance_to_component[source_iid]
```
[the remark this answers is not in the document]
pabloskewes: yes:
cid: component id
iid: instance id
I did it like this to dev faster, but I can rename it

natacha-scopeo (approved):

PR #166, opened 2025-09-25
feat(Bug Alembic): fix alembic revision issue with duplication of tables creation with specific schemas

jccscopeo:
Fix the issue where non public tables are recreated by Alembic when doing new revisions.
Allow Alembic to scan non public schema + change type of enum coming from specific schema to avoid duplication issue when creating a new revision.

To test it, you can pull this branch:
- make db-upgrade
-make db-revision message="test"

and check that test revision indeed has no content

### DIFF
--- a/ada_backend/database/alembic/env.py
+++ b/ada_backend/database/alembic/env.py
@@ -55,6 +55,13 @@ def include_object(object, name, type_, reflected, compare_to):
     if type_ == "table" and name == "apscheduler_jobs":
         return False
 
+    # Exclude tables in test schema
+    if type_ == "table" and str(object.schema) == "test_schema":
+        return False
+    # Exclude test schemas themselves
+    if type_ == "schema" and name == "test_schema":
+        return False
+
     return True  # Include all other objects
 
 


ada_backend/database/models.py
```diff
@@ -35,12 +35,13 @@
 CIPHER = Fernet(settings.FERNET_KEY)
 
 
-def make_pg_enum(enum_cls: Type[StrEnum]) -> SQLAlchemyEnum:
+def make_pg_enum(enum_cls: Type[StrEnum], schema: str = None) -> SQLAlchemyEnum:
```
natacha-scopeo, line 38: enum are link to schema ?

pabloscopeo (approved):
Working locally!

Image

Just this small message too:
```
➜  make db-revision message="test"
Creating a new migration
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.ddl.postgresql] Detected sequence named 'test_table_id_seq' as owned by integer column 'test_table(id)', assuming SERIAL and omitting
  Generating /Users/pabloskewes/Scopeo/draftnrun/ada_backend/database/alembic/versions/58c1f9d532ab_test.py ...  done
```

PR #167, opened 2025-09-25
Marc/dra 454 add the possibility to set global secrets from super admin

marc-scopeo:


engine/llm_services/llm_service.py
```diff
@@ -72,20 +72,20 @@ def __init__(
         self._model_name = model_name
         self._api_key = api_key
         self._base_url = base_url
-        if self._api_key is None or self._base_url is None:
+        if self._api_key is None:
```
natacha-scopeo, line 75: if base_url is None we don't set it ?

PR #170, opened 2025-09-29
feat(models/seed) Add the OpenAI OSS120 and Qwen 32b models served by Cerebras (dra-417)

timotheeredouin:


jccscopeo (changes requested):
The openAI model does not work for the agent if there is no tools attached to the AI agent

natacha-scopeo (approved):

PR #172, opened 2025-09-30
[dra-446] use custom filename in pdf and docx generation tools

natacha-scopeo:


pabloscopeo (approved):
LGTM! Just this [last point](https://github.com/Scopeo/draftnrun/pull/172/files#r2395340504), but it lgtm!

PR #173, opened 2025-10-01
[dra-424] create s3 upload urls endpoint

natacha-scopeo:
# New Endpoint: Generate S3 Upload Presigned URLs

## Endpoint
`POST /organizations/{organization_id}/files/upload-urls`

## Summary
Adds an endpoint to generate presigned S3 URLs so users can upload files directly to S3.


ada_backend/routers/s3_files_router.py
```diff
@@ -5,14 +5,48 @@
...
+            "Failed to generate S3 presigned URLs for organization %s (count=%s)",
+            organization_id,
+            len(upload_file_requests),
+        )
+        raise HTTPException(status_code=500, detail="Internal Server Error") from e
+
+
+# TODO: Refactor endpoint to be more RESTful:
+#       use POST /organizations/{organization_id}/files/upload-urls
+#       instead of /files/{organization_id}/upload
```
marc-scopeo, line 48: this looks done to me

ada_backend/routers/s3_files_router.py
```diff
@@ -39,6 +73,9 @@ async def upload_files(
         raise HTTPException(status_code=500, detail=f"Internal Server Error: {e}")
 
 
+# TODO: Refactor endpoint to be more RESTful:
+#       use POST /organizations/{organization_id}/files/upload-urls
+#       instead of /files/{organization_id}/upload
```
marc-scopeo, line 78: are you sire about that ?

marc-scopeo (approved):

ada_backend/services/s3_files_service.py
```diff
@@ -68,3 +70,53 @@ def delete_file_from_s3(
...
+    organization_id: UUID,
+    upload_file_requests: list[UploadFileRequest],
+) -> list[S3UploadURL]:
+    s3_client = get_s3_client_and_ensure_bucket()
+    upload_urls = []
+    for upload_file_request in upload_file_requests:
+        s3_filename = f"{organization_id}/{uuid4()}_{upload_file_request.filename}"
+        key = sanitize_filename(s3_filename)
+        presigned_url = generate_presigned_upload_url(
+            filename=upload_file_request.filename,
```
jccscopeo, line 114: filename is not a argument of the generate_presigned_upload_url function I think, in my test it fails

ada_backend/routers/s3_files_router.py
pabloscopeo: > this looks done to me

no, it's not done, I think the TODO it's right

ada_backend/routers/s3_files_router.py
pabloscopeo: I agree with the TODO message

ada_backend/services/s3_files_service.py
```diff
@@ -68,3 +70,57 @@ def delete_file_from_s3(
...
+
+def generate_s3_upload_presigned_urls_service(
+    organization_id: UUID,
+    upload_file_requests: list[UploadFileRequest],
+) -> list[S3UploadURL]:
+    s3_client = get_s3_client_and_ensure_bucket()
+    upload_urls = []
+    for upload_file_request in upload_file_requests:
+        s3_filename = f"{organization_id}/{uuid4()}_{upload_file_request.filename}"
+        key = sanitize_filename(s3_filename)
```
jccscopeo, line 112: We need to keep the extension of the file (.pdf), otherwise the ingestion code will fail because it will look for the s3 id of the file and will try to get the extension of the file.

jccscopeo (commented):
Sanitize key should be changed in presigned url like in  s line 63 in s3_routers of upload_files endpoint to keep s3 key with actual extension of the file (.pdf etc). Otherwise Ingestion code will throw an error because it cannot detect the extension from the s3 path

jccscopeo (approved):
Working locally for ingestion , both with s3 bucket and local bucket

PR #176, opened 2025-10-03
feat(Components) Add icons to components (dra-487)

timotheeredouin:
New "icon" parameter in the class Component in models.py
Contains Iconify icon name
Can be retrieved using GET /components/{organization_id} endpoint


ada_backend/database/seed/utils.py
```diff
@@ -87,6 +87,8 @@ def add_custom_llm_model(
     "docx_generation": UUID("b5195a0f-94f5-4f5c-8804-32dd19b16833"),
 }
 
+ICON_NAMES: dict[str] = {"linkup": "streamline-ultimate:network-search"}
```
natacha-scopeo, line 90: No we need to use global variable

PR #529, opened 2026-03-10
feat(scheduler): decouple cron execution from direct calls dra-1056

natacha-scopeo:
# feat(scheduler): decouple cron execution from direct calls

## Summary
Decouples cron job execution from direct in-process agent calls. The scheduler now dispatches HTTP requests to internal API endpoints and returns immediately, instead of calling the agent runner directly and blocking until completion.

### Motivation
- The scheduler should not be aware of internal service implementations. It should only interact with the system through HTTP endpoints, keeping it fully decoupled from the agent execution layer.
- Cron entries were calling `run_env_agent` directly, blocking the scheduler process and holding DB sessions open for the full duration of each run.
- There was no distinct `CallType` for cron-triggered runs, making them indistinguishable from regular API calls in run history.
- The endpoint polling business logic (poll → diff → trigger) was embedded inside the cron entry module, making it hard to test and reuse independently.

_Note : New env variables_
Variable | Description
-- | --
SCHEDULER_API_KEY | Raw key sent by the scheduler in X-Scheduler-API-Key
SCHEDULER_API_KEY_HASHED | Hashed version used for verification

## Summary by CodeRabbit

* **New Features**
  * Internal scheduler endpoint and scheduler API key support; endpoint-polling now dispatches work asynchronously and can trigger workflows per new items.
  * Introduced queued cron lifecycle state and async dispatch result type to support out-of-band run processing.

* **Security / Auth**
  * Unified webhook vs scheduler API-key verification for incoming internal requests.

* **Database**
  * Migrations register new cron-related enum values.

* **Documentation**
  * README adds scheduler API key generation and example credentials.

* **Tests**
  * Endpoint polling tests updated for async dispatch behavior.


pabloscopeo (commented):
Left some questions! Plus:
- Before, cron_id was propagated like this: `run_env_agent(..., cron_id=cron_id)` I don't remember if we fixed this to be done via context vars, but if we dind't then we would be losing that for logs. Can we check?
- we probably need to add the new envars to example?
- if we are doing this via http maybe let's check if we don't have unnecessary db sessions on scheduler/service.py

pabloscopeo (approved):
LGTM, it would be nice to add a TODO to rename the endpoint (as previous message)

PR #575, opened 2026-03-24
feat(FieldExpression): LLM as a field expression DRA-1105

dianascopeo:
## Summary by CodeRabbit

* **New Features**
  * Added selectable completion-model inputs across many components and tools so users can choose a model per component; UI model options improved.

* **Database**
  * Migration to convert existing completion-model parameter storage to the new port/field-expression format to preserve selections.
  * Updated seed data to stop creating a separate completion-model parameter.

* **Documentation**
  * Added guidance on build-time vs. runtime resolution of model-selection field expressions.


pabloskewes (commented):
This it's for migrating LLM Model's name as FE instead of static param right? if yes it lgtm but why is there component migration too?

pabloscopeo (commented):
Using LLM as Input it's a nice addition :)

I think we might be able to simplify the service layer a bit though. The `BUILD_TIME_PARAMETER_TYPES` resolution and the `CompletionServiceFactory` pattern add a lot of machinery, but `completion_model` could just flow as a normal runtime input and be consumed where it's used. The idea of supporting a system like this it's good I think, but I think that maybe it's better to not address it like this now. Do you think we can do something like this? (approx):

```python
# LLMCall / AIAgent: build the service inside _run_without_io_trace
async def _run_without_io_trace(self, inputs: LLMCallInputs, ctx: dict) -> LLMCallOutputs:
    provider, model_name = get_llm_provider_and_model(inputs.completion_model)
    completion_service = CompletionService(provider=provider, model_name=model_name, ...)
    ...

# RAG: pass it down to the synthesizer
sourced_response = await self._synthesizer.get_response(
    ...,
    completion_model=inputs.completion_model,
)

# Synthesizer.get_response: build inline
async def get_response(self, ..., completion_model: str, ...):
    provider, model_name = get_llm_provider_and_model(completion_model)
    completion_service = CompletionService(...)
    ...
```

This would let us drop `BUILD_TIME_PARAMETER_TYPES`, `get_build_time_port_defaults`, and the factory processor entirely. The tradeoff is building `CompletionService` per-run instead of once at init, but that's fine as complexity wise, we still do 1 instantiation = 1 run. (Unless there's some issue I'm not seeing).

Happy to chat through it! :)

pabloscopeo (commented):
LGTM! Just a few questions so I get 100% of the PR:
- what are the changes en get graph service for?
- same for update service, why do we need to modify the services?

pabloscopeo (approved):
LGTM! Just, before merging, can we please:
- Move the new repo function outside the port mappings repo module, as that one will disappear
- Add TODOs on the logic added in `get_graph_service` and `uddate_pipeline_service` that way that we should remove input-specific logic from those services when we support a way to allow dynamic data on ui components (ex: fetching slack channels).

pabloscopeo (approved):
LGTM!

ada_backend/services/components_service.py
```diff
@@ -70,6 +72,12 @@ def _process_components_with_ports(
         component.parameters = filter_conflicting_parameters(component.parameters or [], input_ports)
 
         for input_port in input_ports:
+            props = input_port.ui_component_properties
+            if input_port.parameter_type == ParameterType.LLM_MODEL:
+                caps = (props or {}).get("model_capabilities")
```
natacha-scopeo: naming

ada_backend/services/components_service.py
```diff
@@ -70,6 +72,12 @@ def _process_components_with_ports(
...
+                props = get_ui_component_properties_with_llm_options(
```
natacha-scopeo: props => naming

PR #619, opened 2026-04-03
feat(component): fix coercion error dra-1195

natacha-scopeo:
# fix: coerce JSON strings to dict/list for component inputs (DRA-1195)

## Summary
- Fix "cannot coerce" error when `knowledge_search_retriever` (or any component) receives JSON-encoded strings (`"{}"`, `'["a"]'`) for dict or list typed input ports — a common scenario when values originate from a LiteralNode or LLM tool call.
Add `str` → `dict`, `str `→ `list` (and their `Optional` variants) coercion rules to the CoercionMatrix, and apply the coercion matrix to the tool-invocation path in Component before Pydantic validation.

## Summary by CodeRabbit

* **New Features**
  * Components now auto-parse JSON-encoded strings into structured types (dicts and lists) for inputs and filters, enabling passing JSON strings for queries/filters.

* **Tests**
  * Added tests covering JSON string coercion, preprocessing of inputs, and retriever behavior with string filters.

### DIFF
--- a/engine/coercion_matrix.py
+++ b/engine/coercion_matrix.py
@@ -22,6 +22,7 @@ GRAPH_RUNTIME_TARGET_TYPES: frozenset[type] = frozenset({
     float,
     bool,
     dict,
+    list,
     list[str],
     list[dict],
     list[ChatMessage],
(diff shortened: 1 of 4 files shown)


pabloscopeo (approved):
LGTM, especially if this is legacy only, as it will be not necessary pretty soon

pabloscopeo (approved):
After knowing that this is not a legacy path, I think it's still fine.

I wonder if pydantic performs this coercion correctly, which would imply that the only problem we had, it's that this was crashing with CoercionError before arriving to pydantic, which means that maybe we need to fix that instead.

Ideally, we shouldn't need the CoercionMatrix anymore after cleaning the legacy. This is a simple run time glue we had to add when migrated to typed I/Os, I'm not 100% sure rn if we will be able to delete it the moment we clean, but to me, ideally, it would be best to rely on pydantic instead.

PR #631, opened 2026-04-08
Remove port mappings - DRA-1120

pabloscopeo:
## What changed

This PR removes the legacy `port_mappings` wiring path and makes field expressions the single source of truth for graph wiring.

### Backend / engine
- Removed `port_mappings` usage from graph update/get/deploy services and `agent_runner_service`.
- Simplified `GraphRunner` to expression-only input gathering (no mapping synthesis/validation path).
- Removed `PortMapping` ORM/schema/repository surface (`models.py`, `port_mapping_schema.py`, admin, MCP docs/tools updates).
- Kept canonical wiring behavior via auto-generated field expressions.

### Migration
- Added `g6h7i8j9k0l1_drop_port_mappings_table.py` (`code-first`):
  - defensive backfill for any mapping lacking a corresponding field expression
  - drop `port_mappings` table
  - downgrade recreates table schema (empty), so rollback remains operational with expression fallback

### Tests/docs
- Reworked/removing tests tied to legacy mapping semantics; adapted engine/backend/MCP tests to expression-only behavior.
- Fixed regression where top-level `RefNode` values were stringified (now preserved as raw object).
- Updated `ada_backend/docs/engine.md` and Cursor rules to reflect expression-only architecture.

## Why this solution

`port_mappings` duplicated wiring already represented by field expressions and created parallel behavior in save/load/runtime flows.  
Removing it reduces execution complexity, removes redundant persistence, and aligns UI/backend/engine around one canonical wiring model.

## Out of scope

- Legacy `AgentPayload`/`migrated` compatibility cleanup is not part of this PR, as well as legacy paths.

## Validation

- Local test runs after adaptation: **1306+ passing** with remaining failures resolved in subsequent fixes for this branch.
- Migration verified locally end-to-end:
  - pre-upgrade: `port_mappings` present (161 rows, 6 missing expressions)
  - upgrade: missing expressions backfilled, `port_mappings` dropped
  - downgrade: table recreated with expected schema and 0 rows
  - re-upgrade: successful, DB ends at migration head
- MCP local functional QA (expression-only wiring):
  - verified canonical edge auto-wiring still creates `RefNode` field expressions (no `port_mappings` path)
  - verified draft lifecycle (`save_graph_version` + `publish_to_production`) preserves wiring across runner cloning/remapped instance IDs
  - verified runtime execution on both production runner and new draft runner after publish
  - verified `update_component_parameters` round-trip preserves existing field expressions
  - verified `get_graph` no longer returns `port_mappings`
  - verified expression/runtime scenarios in real workflows: `ref`, `concat`, `json_build`, key-extraction path, If/Else routing, diamond topology, and agent-tool relationships
- Note on one MCP scenario: `output_format` + key extraction depended on model structured-output behavior (non-deterministic); expression evaluation path behaved as expected and no regression was attributed to `port_mappings` removal.


engine/graph_runner/field_expression_management.py
```diff
@@ -122,6 +122,11 @@ def evaluate_node(node: ExpressionNode) -> str:
             LOGGER.debug(f"Evaluated JSON build expression for {target_field_name}")
             return result
 
+        case RefNode() as ref:
```
natacha-scopeo, line 130: why adding this ?

engine/graph_runner/graph_runner.py
```diff
@@ -312,146 +280,23 @@ def _to_string(value: Any) -> str:
...
-                if target_id == node_id and not isinstance(expr_ast, RefNode)
-            ]
-
-            target_component = self.runnables[node_id]
-            for field_name, expression_ast in non_ref_expressions:
-                LOGGER.debug(f"Evaluating non-ref expression for {node_id}.{field_name}")
-                self._eval_expression_and_set_input(
-                    node_id, field_name, expression_ast, target_component, input_data, log_prefix="expression"
-                )
+        if is_start_node and not real_predecessors:
```
natacha-scopeo: no Diana did a fix for this. if not a link to a start node the component should not be trigger. @dianascopeo what was you fix exactly ?

ada_backend/services/graph/update_graph_service.py
```diff
@@ -556,84 +536,24 @@ async def update_graph_service(
...
 
-def _ensure_port_mappings_for_edges(  # TODO(port-mapping-removal): delete when port mappings are removed
+def _ensure_canonical_expressions_for_edges(
     session: Session,
     graph_runner_id: UUID,
     graph_project: GraphUpdateSchema,
     input_params_by_instance: dict[UUID, list] | None = None,
 ) -> list[AutoGeneratedCanonicalExpression]:
-    """Ensure explicit port mappings exist for all edges.
+    """Auto-generate canonical RefNode field expressions for edges that have no explicit wiring.
```
natacha-scopeo, line 545: can you explain this ? what do you check ? what is a ref node ?

engine/graph_runner/graph_runner.py
dianascopeo: my solution was when we have start node we will run only that graph but for old projects i have a todo if i am not mistaken

pabloscopeo: I took a look at https://github.com/Scopeo/draftnrun/pull/532, and they are different concerns, complementary:

- Diana -> Assures only nodes reachable from a trigger enter the graph (before GraphRunner is built)
- This check -> Assures the graph entrypoint receives the user's input payload as initial data (inside the engine). Without it, the entrypoint runs with empty input_data {} so messages never reaches the first component.

Also renamed is_start_node -> is_graph_entrypoint here to avoid confusion with the Start component concept, it's a topological property, not a component type check.

engine/graph_runner/field_expression_management.py
pabloscopeo: This fixes a pre-existing bug that was hidden by port mappings. In main, canonical PM wiring went through the port mappings path in `_gather_inputs`, which passed values as raw Python objects. The field expressions path also existed but was rarely the sole route, so a top-level `RefNode` falling through to `evaluate_node()` -> `to_string()` never caused visible issues.

With port mappings removed, all wiring goes through field expressions exclusively. Now a plain `ref(start, "messages")` hits the `case _:` fallback -> `evaluate_node()` -> `to_string(raw_value)`, turning a `list[ChatMessage]` into its string repr `"[{'role': 'user', ...}]"`. The downstream component receives a string instead of a structured list and breaks.

The fix is adding the explicit `case RefNode()` top-level branch that returns the raw object, same as the `case VarNode()` that was already there for the same reason.

ada_backend/services/graph/update_graph_service.py
pabloscopeo: Same function as `_ensure_port_mappings_for_edges` on main, it used to write a PortMapping row and an auto-generated RefNode expression per edge. Now that port mappings are gone, it only writes the expression. The auto-generated part is unchanged.

engine/graph_runner/graph_runner.py
```diff
@@ -312,146 +280,23 @@ def _to_string(value: Any) -> str:
...
-                        field_name,
-                        e,
-                    )
+        if is_graph_entrypoint and not real_predecessors:
+            input_task_result = self.tasks[self._input_node_id].result
+            input_data = dict(input_task_result.data) if input_task_result else {}
+
+        target_component = self.runnables.get(node_id)
+        for field_name, expr_ast in self._expressions_by_node.get(node_id, []):
+            self._eval_expression_and_set_input(node_id, field_name, expr_ast, target_component, input_data)
```
natacha-scopeo, line 299: I think I still do get the entrypoint concept

pabloscopeo: GraphRunner adds a virtual `__input__` node pointing to all nodes with no real predecessors. is_graph_entrypoint = `__input__` is in my predecessors = I'm a root node with no upstream components = I should receive the initial user payload."

natacha-scopeo (approved):
need big big QA

PR #637, opened 2026-04-10
feat(tool): updategoogle agenda tool description dra-1213

natacha-scopeo:
# Add calendar_get_my_email tool and improve attendee guidance for Google Calendar

## Summary
- Added a new `calendar_get_my_email `MCP tool that returns the authenticated calendar owner's email address, so agents can look it up when building attendee lists.
- Updated the `calendar_create_event` tool description to remind the LLM to include the calendar owner as an attendee when they should participate, without forcing it for every event.
- Added `get_user_email()` method to `GoogleCalendarClient` with caching to avoid redundant API calls.

### Context
When a user asked their agent to "add a meeting with Marc to my agenda", the agent would create the event with only Marc as an attendee, leaving the calendar owner out of the meeting. Rather than silently injecting the owner server-side (which would break use cases like scheduling a meeting between other people), we expose the owner's email as a tool and guide the LLM via the tool description.

## Release Notes

* **New Features**
  * Introduced a new tool to retrieve your primary calendar email address, simplifying event management workflows.

* **Documentation**
  * Enhanced the event creation tool description with guidance on properly adding the calendar owner when inviting attendees.

* **Tests**
  * Added tests to validate the new email retrieval functionality works as expected.

### DIFF
--- a/engine/components/tools/google_calendar_mcp/client.py
+++ b/engine/components/tools/google_calendar_mcp/client.py
@@ -22,6 +22,16 @@ class GoogleCalendarClient:
             raise ValueError("access_token is required")
         creds = Credentials(token=access_token)
         self._service = build("calendar", "v3", credentials=creds)
+        self._user_email: str | None = None
+
+    async def get_user_email(self) -> str:
+        if self._user_email is None:
+            def _call():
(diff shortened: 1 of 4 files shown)


dianascopeo (approved):
you do not need to modify mcp as well?

PR #642, opened 2026-04-13
refactor(tracing): decouple cron_id from agent_runner via span_context merge DRA-1003

pabloscopeo:
## What changed

`agent_runner_service` no longer knows `cron_id` exists. Previously `cron_id` was threaded as an explicit parameter all the way down: `executor → run_env_agent → run_agent → setup_tracing_context → set_tracing_span`.

The fix is a merge-capable `set_tracing_span`: cron executors set `cron_id` in the tracing context before calling `run_env_agent`, and when `run_agent` later calls `set_tracing_span` with project/org info it only patches those fields — `cron_id` is preserved automatically via `dataclasses.replace`.

Additionally (because execution now goes through internal webhooks), this PR propagates `cron_id` across the HTTP boundary:
- cron executor now sends both `cron_id` (job id) and `cron_run_id` (execution id) in the internal webhook body
- API-side `_execute_cron_run` sets `set_tracing_span(cron_id=...)` before calling `run_env_agent`
- `RunProjectBody` now accepts `cron_id`

This keeps `agent_runner_service` decoupled while preserving cron correlation in API-side traces.

## Usage

`set_tracing_span(**kwargs)` merges with the existing context — it only overwrites the fields you pass, everything else is preserved. Call it as many times as you want from different layers:

```python
set_tracing_span(cron_id="abc-123")           # upstream, before run_env_agent
set_tracing_span(project_id="proj", ...)      # downstream, inside run_agent
# cron_id is still "abc-123" after the second call
```

Note: passing `None` explicitly **does** overwrite. Only omitting a field preserves it.

For webhook-triggered cron execution:
- `cron_id` = cron job id
- `cron_run_id` = specific cron execution id

## Testing

Validated locally via MCP:
1. Created a cron job (`agent_inference` entrypoint) via `create_cron`
2. Triggered it manually via `trigger_cron`
3. Confirmed `[cron_id=...]` correctly propagates to all logs in the execution path (`ada_backend.services.cron.entries.agent_inference`, `httpx`, `ada_backend.services.cron.execution`) without passing through `agent_runner_service`.

Also validated with automated tests:
- `tests/engine/trace/test_span_context.py`
- `tests/ada_backend/routers/test_webhook_internal_router.py`

## Summary by CodeRabbit

* **Improvements**
  * Tracing and cron correlation strengthened so scheduler-triggered runs propagate a cron identifier across internal webhook and background run boundaries.

* **Documentation**
  * Webhook API docs and architecture guidance updated to require/describe cron_id (and cron_run_id semantics) for scheduler-triggered flows.

* **Tests**
  * New tests added to verify tracing behavior and that cron_id is forwarded to downstream/child workflow runs.


ada_backend/services/cron/entries/agent_inference.py
natacha-scopeo, line 134: is this solution going to work now that we use the webhook ?

engine/trace/span_context.py
```diff
@@ -26,32 +27,23 @@ class TracingSpanParams:
...
+
+def set_tracing_span(**kwargs) -> None:
+    """Set or merge tracing context fields.
+
+    When a context already exists, only the provided fields are updated;
+    all other fields (e.g. cron_id set upstream) are preserved.
+    """
+    invalid = set(kwargs) - _SPAN_FIELDS
+    if invalid:
+        raise TypeError(f"Unknown tracing span fields: {invalid}")
```
natacha-scopeo, line 45: add logger

natacha-scopeo, line 45: and why raising an error instead of ignoring it ?

pabloscopeo: if we do `set_tracing_span(inexistant_arg="...")` it should crash, ignoring it would be more confusing, the dev wouldn't get any signal to know that the value it's not being used

ada_backend/services/cron/entries/agent_inference.py
pabloscopeo: The cron_id actually stopped reaching the agent traces when we migrated to webhooks. This `set_tracing_span` here only covers the scheduler process.

To get the `cron_id` back into the agent's traces on the API side, we need to pass it in the webhook payload (maybe as header? same as `cron_run_id`) and set it in `_execute_cron_run`. I can add that fix to this PR if you want.

engine/trace/span_context.py
pabloscopeo: added logs (though maybe it wasn't that necessary as we have the error itself, but why not)

ada_backend/services/cron/entries/agent_inference.py
natacha-scopeo: yes we need this information inside the run

pabloscopeo: done

natacha-scopeo (approved):
LGTM need to test client cron and webhook

natacha-scopeo (approved):

PR #648, opened 2026-04-14
Fix RefNode expr stringifying list values passed to components dra-1223

natacha-scopeo:
# Fix RefNode expression stringifying list values passed to downstream components

## Summary
- Root cause fix: Top-level RefNode expressions in `evaluate_expression()` fell through to the default `_` case, which called `evaluate_node()` — a function that always returns `str.` This turned list values like` ["file1.pdf", "file2.docx"]` into the string "`['file1.pdf', 'file2.docx']"`, which coercion then wrapped as `["['file1.pdf', 'file2.docx']"]`. Added an explicit `RefNode` case that calls `evaluate_ref_as_object()` to preserve the raw value (list, dict, etc.), matching how `VarNode` and `JsonBuildNode` were already handled.
- Defense-in-depth: Added `field_validator("email_attachments")` on `GmailSenderInputs` and `OutlookSenderInputs` to flatten nested lists and wrap bare strings, mirroring the existing email_recipients validator.

## Context
Users wiring file-producing components (e.g. DOCX template, PDF generation) into Gmail/Outlook via field expressions hit:
```
FileNotFoundError: Attachment not found or not a file:
  089e5d94-…/['Rapport_Pre-Instruction_Kheoos_2026-04-14.pdf', 'kheoos_PreInstruction_filled.docx']
```
The stringified Python list was concatenated as a single filename instead of being iterated as individual attachment paths.

## Summary by CodeRabbit

* **Improvements**
  * Email attachment inputs now support flexible formats and are automatically normalized across email integrations for more consistent processing
  * Expression evaluation capabilities have been enhanced to support reference expressions at the top level, providing greater flexibility for complex expression scenarios

### DIFF
--- a/engine/integrations/gmail/gmail_sender.py
+++ b/engine/integrations/gmail/gmail_sender.py
@@ -12,7 +12,12 @@ from ada_backend.database.setup_db import get_db_session
 from engine.components.component import Component
 from engine.components.types import ComponentAttributes, ToolDescription
 from engine.integrations.gmail.gmail_utils import create_raw_mail_message
-from engine.integrations.utils import get_gmail_sender_service, get_google_user_email, get_oauth_access_token
+from engine.integrations.utils import (
+    get_gmail_sender_service,
+    get_google_user_email,
+    get_oauth_access_token,
+    normalize_str_list,
+)
 from engine.trace.trace_manager import TraceManager
 from settings import settings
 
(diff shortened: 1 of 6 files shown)


engine/integrations/gmail/gmail_sender.py
```diff
@@ -145,6 +145,21 @@ def validate_email_recipients(cls, v):
...
+        if isinstance(v, list):
+            flattened: list[str] = []
+            for item in v:
+                if isinstance(item, list):
+                    flattened.extend(str(i) for i in item)
+                else:
+                    flattened.append(str(item))
+            return flattened
+        return v
+
```
dianascopeo, line 157: better to be in utils becasue outlook and gmail share same code

dianascopeo (approved):

PR #650, opened 2026-04-14
fix(security): plaintext secrets in logs - DRA-1254

pabloscopeo:
## What changed

This PR implements a defense-in-depth strategy to prevent secrets (API keys, tokens, passwords) from leaking into logs, traces, and Sentry. The changes are grouped into 4 main architectural approaches:

**1. In-Memory Protection (`SecretStr` & Late Unwrap)**

- Adopted `pydantic.SecretStr` as the canonical type for secrets read from the database (e.g., `OrganizationSecretDTO`).
- This ensures that if a secret object is accidentally logged or stringified before processing, it renders as `**********`.
- Secrets are explicitly unwrapped to plaintext (`.get_secret_value()` / `unwrap_secret()`) only at the final execution boundaries (e.g., placeholder resolution, component instantiation, SDK clients).

**2. Logging Hygiene (Log Structure, Not Values)**

- Eradicated the practice of logging full parameter dictionaries or attempting to manually redact specific keys (which is brittle).
- **Services:** In `agent_builder_service.py` and `entity_factory.py`, error logs now only emit parameter names (`list(input_params.keys())`) instead of the full resolved payload.
- **Clients:** In `redis_client.py`, removed hardcoded redaction logic (`if k == "access_token"`) in favor of purely structural logging (`keys={list(payload.keys())}`).

**3. Best-Effort Heuristic Redaction**

- Introduced `engine/log_redaction.py` to catch plaintext secrets that bypass the `SecretStr` typing (e.g., external inputs, MCP tool arguments).
- Uses key-name matching (`api_key`, `token`, `password`) and regex (`Bearer ...`) to mask sensitive data recursively.
- Applied as a global `before_send` hook in the backend and scheduler Sentry processes, and integrated directly into `serialize_to_json` to protect OpenInference traces. The Redis worker scrubber was split out into follow-up `DRA-1263` to avoid a `worker -> ada_backend` dependency.

**4. Dual Prompt Rendering (`reveal_secrets`)**

- Added a `reveal_secrets` flag to `fill_prompt_template`.
- Components like `ai_agent` and `llm_call` now render prompts twice: once with real secrets for actual LLM execution, and once with masked secrets to safely record the interaction in the tracing system (`SpanAttributes.LLM_INPUT_MESSAGES`).

## Why

Secrets were flowing as plain strings through the codebase, making them vulnerable to accidental exposure via `LOGGER.debug/error`, Sentry exception payloads, or OpenInference trace spans. A single strategy wasn't enough: we needed strong typing (`SecretStr`) for known secrets, strict logging hygiene for internal payloads, heuristic redaction for untyped external data, and dual-rendering for LLM observability.

### Follow up

- **Embedded Placeholders:** Currently, `secret_resolver.py` unwraps to plain `str` immediately. Embedded placeholders (e.g., `"Bearer @{ENV:TOKEN}"`) lose their `SecretStr` protection early. Propagating a hybrid concatenable wrapper requires refactoring all downstream consumers.
- **Redis Worker Sentry Scrubber:** The worker-specific Sentry scrubber was removed from this PR after review because the worker must not import from `ada_backend`. That reintroduction is tracked separately in `DRA-1263`.
- **Redis Worker Shared Module / Dependency Cleanup:** Any shared-helper extraction or other architecture needed to restore worker Sentry scrubbing without a `worker -> ada_backend` dependency is intentionally left to `DRA-1263`, not this PR.

### Eventually

- **Queue Persistence:** Raw payloads serialized to Redis streams (`client.xadd`) or PostgreSQL (`run_inputs`) still contain plaintext secrets. Encryption at rest for these queues is a larger architectural problem deferred to a future PR.

## Validation

- Added unit tests for `log_redaction.py` (key matching, regex, nested structures).
- Updated `test_secret_resolver.py` to verify unwrap behavior.
- Added tests for `serialize_to_json` heuristic redaction and `api_call_tool` error masking.
- All tests pass locally.


engine/graph_runner/field_expression_management.py
```diff
@@ -127,12 +134,13 @@ def evaluate_node(node: ExpressionNode) -> str:
             LOGGER.debug(f"Evaluated JSON build expression for {target_field_name}")
             return result
 
+        # TODO: duplicate of the RefNode case above, never reached — remove.
         case RefNode() as ref:
             result = evaluate_ref_as_object(ref)
-            LOGGER.debug(f"Evaluated ref expression for {target_field_name}: {result}")
+            LOGGER.debug("Evaluated ref expression for %s (type=%s)", target_field_name, type(result).__name__)
             return result
```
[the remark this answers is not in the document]
pabloscopeo: why on this PR? I put a TODO to keep track of it, but it's completely out of scope

ada_backend/services/trace_service.py
```diff
@@ -219,8 +223,13 @@ def get_root_traces_by_project(
         end_time=end_time,
     )
     track_monitoring_loaded(user_id, project_count=1, organization_id=organization_id)
-    LOGGER.info("Querying root spans for project %s with duration=%s start_time=%s end_time=%s",
-                project_id, duration, start_time, end_time)
+    LOGGER.info(
```
natacha-scopeo: debug

ada_backend/services/entity_factory.py
```diff
@@ -881,6 +885,7 @@ def processor(params: dict, constructor_params: dict[str, Any]) -> dict:
         engine_url = params.pop("engine_url", None)
         if not engine_url:
             return params
+        engine_url = unwrap_secret(engine_url)
```
natacha-scopeo: why adding this ? is engine_url save at a secret ? (i don't think so)

ada_backend/utils/redis_client.py
```diff
@@ -397,6 +398,9 @@ def push_run_task(
         LOGGER.error("Redis client unavailable. Cannot push run task %s", run_id)
         return False
 
+    # TODO(security): `input_data` is persisted verbatim on the Redis runs queue and
+    # mirrors whatever the caller passed (including any inlined secrets). Sanitization
+    # here requires a convention for sensitive fields or per-graph metadata.
```
natacha-scopeo, line 405: this I don't get. this is the input of the run (it is not really related to this PR)

engine/components/ai_agent.py
```diff
@@ -333,11 +336,11 @@ async def _run_tool_call(
         # Merge pre-configured literal inputs with LLM-provided arguments; LLM args win.
         merged_arguments = {**pre_configured, **tool_arguments}
         try:
-            LOGGER.info(f"Calling tool {tool_function_name} with arguments: {merged_arguments}")
+            LOGGER.info(f"Calling tool {tool_function_name} with argument keys={list(merged_arguments.keys())}")
```
natacha-scopeo: debug

engine/components/ai_agent.py
```diff
@@ -333,11 +336,11 @@ async def _run_tool_call(
...
             tool_output = await tool_to_use.run(ctx=ctx, **merged_arguments)
-            LOGGER.info(f"Tool {tool_function_name} returned: {tool_output}")
+            LOGGER.info(f"Tool {tool_function_name} returned (type={type(tool_output).__name__})")
```
natacha-scopeo: What is {type(tool_output).__name__} ? and why only type ? We don't get information with it

natacha-scopeo: debug

engine/components/ai_agent.py
```diff
@@ -456,9 +465,17 @@ async def _run_core(
         tool_choice = "auto" if self._current_iteration + 1 < self._max_iterations else "none"
         with self.trace_manager.start_span("Agentic reflexion") as span:
             llm_input_messages = [msg.model_dump() for msg in history_messages_handled]
+            trace_input_messages = [dict(message) for message in llm_input_messages]
+            for message in trace_input_messages:
+                if message.get("role") == "system":
+                    message["content"] = masked_system_prompt
+                    break
```
natacha-scopeo, line 471: I found this really complex just for masking a system prompt. I don't like the iteration on message + the if condition

engine/components/llm_call.py
```diff
@@ -204,9 +205,11 @@ def _extract_file_url(self, inputs: LLMCallInputs, ctx: Optional[dict]) -> list:
         return []
 
     async def _run_without_io_trace(self, inputs: LLMCallInputs, ctx: Optional[dict] = None) -> LLMCallOutputs:
-        LOGGER.info(f"Running LLM call agent with inputs: {inputs} and ctx: {ctx}")
-        prompt_template = inputs.prompt_template
+        LOGGER.info(f"Running LLM call agent for {self.component_attributes.component_instance_name}")
+        prompt_template = inputs.prompt_template or ""
```
natacha-scopeo: why adding this or ? it change the logic of the component

natacha-scopeo (commented):
Quite hard to review...To make review easier, I would have preferred to have it in several PR.

engine/components/llm_call.py
pabloscopeo: this is a good catch actually, I didn't see it

ada_backend/utils/redis_client.py
pabloscopeo: > not really related to this PR

Exactly, that's why it's a `TODO` and not a fix. I flagged it during the audit because it's the same class of issue as the `run_inputs` row in Postgres (also called out in the PR description under *Known Limitations*): whatever the caller passes as `input_data` gets persisted verbatim.

The risk is inlined secrets, anything the caller put into the payload directly (user-pasted tokens, etc.). Redis and PG end up with plaintext. Fixing it properly needs a deeper solution, which is out of scope here. I'm leaving the TODO so the next person who goes to log/forward/export this payload sees the liability before doing it.

engine/components/ai_agent.py
pabloscopeo: ok no prob (btw, I maintained the debug level on every log change, I don't mind changing it but just so you know)

engine/components/ai_agent.py
pabloscopeo: true, it's a useless leftover (left after applying this pattern on all logs, but here it's doesn't add info, will delete)

ada_backend/services/entity_factory.py
pabloscopeo: you're right, checked the seed and `engine_url` is STRING, not a secret. added it defensively because imo it should be a secret (the placeholder is literally `postgresql://user:password@host/db`), but it's a no-op today. can drop it if you prefer.

engine/components/ai_agent.py
pabloscopeo: yeah I agree it doesn't look pretty, but it's what makes more sense, as the masking (via fill_template function) goes only for the system prompt. I can put it on a helper if you want, to hide the logic.

ada_backend/services/trace_service.py
pabloscopeo: done

engine/components/ai_agent.py
pabloscopeo: done

engine/components/ai_agent.py
pabloscopeo: deleted

pabloscopeo: done

engine/components/llm_call.py
pabloscopeo: reverted

pabloscopeo:
> Quite hard to review...To make review easier, I would have preferred to have it in several PR.

Oh, I'm sorry for that. In my head it wasn't hard because there's 4 main things (explained on the PR description), where (1) it's mandatory (main security barrier) and the diff it's simply wrapping and unwrapping strings, then (3) it's like, best effort for redaction when we have no choice (kinda necessary on the same PR too, and diff-wise it's a new module + calling it at some places), and finally (2) that it's simply reducing log info.

I do agree that (4) could have stayed a TODO on this PR, but I wanted to add it as messages are sent to traces, and secrets are necessarily resolved on prompts (when injected), so it was hard for me to leave it out.

I will take note. Also, I can still split now if you want, but if you ask me I think it's not that necessary as the changes are kinda low risk imo.

ada_backend/services/entity_factory.py
natacha-scopeo: yes we need to really fix the pb

pabloscopeo: dropped and added TODO

ada_backend/services/trace_service.py
```diff
@@ -45,7 +45,11 @@ def get_attributes_with_messages(span_kind: str, row: pd.Series, filter_to_last_
         if not input:
             input = input_data
     except Exception as e:
-        LOGGER.error(f"Error extracting last user message: {e}, input_data: {input_data}")
+        LOGGER.error(
+            f"Error extracting last user message: {type(e).__name__} "
+            f"(input_data type={type(input_data).__name__}, "
```
natacha-scopeo: still the type stuff

engine/graph_runner/graph_runner.py
```diff
@@ -279,7 +282,7 @@ def _to_string(value: Any) -> str:
                 )
                 evaluated_value = coerced
         input_data[field_name] = evaluated_value
-        LOGGER.debug(f"Set {node_id}.{field_name} from {log_prefix}: {evaluated_value}")
+        LOGGER.debug(f"Set {node_id}.{field_name} from {log_prefix} (type={type(evaluated_value).__name__})")
```
natacha-scopeo: still the type suff

natacha-scopeo (approved):

workers/worker/base_worker.py
```diff
@@ -13,6 +13,7 @@
 import sentry_sdk
 from dotenv import load_dotenv
 
+from ada_backend.utils.log_redaction import scrub_sentry_event
```
natacha-scopeo: no

natacha-scopeo (changes requested):

workers/worker/base_worker.py
pabloscopeo: fixed

.cursor/rules/architecture.mdc
```diff
@@ -49,6 +49,7 @@ Free-form string tags on projects via `project_tags` table (many-to-many). Tags
...
 - Ingestion worker: Redis Stream consumer, separate process.
 - Scheduler: APScheduler + PostgreSQL job store, separate process. Registry-based cron types in `ada_backend/services/cron/`. See `ada_backend/services/cron/Readme.md`.
+- Shared code used by both ingestion and webhook workers must be safe for the webhook worker image. In practice, modules imported by `workers/worker/base_worker.py` cannot depend on `engine/` because `Dockerfile.webhook-worker` does not copy that package. Temporary cross-runtime utilities like log redaction may live in `ada_backend/utils/` until they are extracted to a dedicated shared module that all relevant images include.
```
natacha-scopeo: not up to date ?

pabloscopeo: yes, fixed now

natacha-scopeo (approved):

PR #651, opened 2026-04-14
feat(ci): add integration tests dra-1095

natacha-scopeo:
# Add E2E integration tests for project flow (save / deploy / run)

## Summary
- Add a pytest E2E suite (`scripts/test_project_flow.py`) that exercises the critical save-version -> deploy -> run path against staging for a configurable list of projects.
- Add a manually-triggered GitHub Actions workflow (`test-project-flow.yml`) to run the suite on demand via workflow_dispatch.

## Motivation
We have no automated way to verify that the core project lifecycle works end-to-end on staging after a deploy. This gives us a one-click smoke test that can catch regressions before they reach production.

## How it works
- The test authenticates against Supabase with a dedicated test user (email + password) to obtain a real JWT — no auth bypass or backend code changes needed.
- For each project in the `PROJECT_TEST_MATRIX`, three tests run sequentially: `test_save_version`, `test_deploy, test_run`.
Results are published as a JUnit report artifact.

### DIFF
--- a/.github/workflows/test-project-flow.yml
+++ b/.github/workflows/test-project-flow.yml
@@ -1,0 +1,50 @@
+name: E2E Project Flow Test
+
+on:
+  workflow_dispatch:
+
+jobs:
+  test-project-flow:
+    runs-on: ubuntu-latest
+
+    env:
+      STAGING_BASE_URL: ${{ secrets.STAGING_BASE_URL }}
+      SUPABASE_URL: ${{ secrets.E2E_SUPABASE_URL }}
+      SUPABASE_ANON_KEY: ${{ secrets.E2E_SUPABASE_ANON_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test-project-flow.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install dependencies
+        run: pip install requests pytest
```
pabloskewes, line 28: maybe we could use some cached version to win some time

pabloskewes (approved):
why on /scripts instead of on /tests with a mark?

pabloscopeo (approved):

PR #652, opened 2026-04-14
feat(tracing): propagate span context to sentry tags

pabloscopeo:
## What changed
Added a generic bridge that automatically propagates `TracingSpanParams` fields to Sentry tags whenever the tracing span context is updated via `set_tracing_span`.

## Why
Currently, rich per-request context (`cron_id`, `trace_id`, `project_id`, `organization_id`, etc.) is available in `TracingSpanParams` and reaches logs and OTel spans, but none of it reaches Sentry. This makes Sentry events and transactions unsearchable by business-domain context (like `cron_id` or `project_id`).

## Implementation Context
- **Whitelist approach:** Defined `SENTRY_TAG_FIELDS` in `engine/trace/span_context.py` to ensure only meaningful scalar fields are propagated (excluding complex objects like `shared_sandbox` or lists).
- **Isolation Scopes:** Leverages `sentry_sdk.get_isolation_scope()` to set/remove tags. This ensures tags are correctly scoped to the current async task/request in FastAPI and don't leak across requests.
- **No-op safety:** `sentry_sdk.set_tag` is a no-op when the SDK is not initialized, making it safe to call unconditionally across all processes (API, scheduler, MCP, ingestion).
- **Stale tag prevention:** If a field is explicitly set to `None`, it is removed from the Sentry isolation scope via `remove_tag`.
- **Docs:** Updated `ada_backend/docs/engine.md` and `.cursor/rules/architecture.mdc` to document this as the canonical pattern.

## Out of Scope
- Broad Sentry observability redesign beyond tracing-context tag propagation.
- Any additional cron propagation changes (already covered by DRA-1003).

## Validation / Testing
- Added regression tests in `tests/engine/trace/test_span_context.py` to verify:
  - Sentry tag sync for whitelisted fields.
  - `None` handling (ensuring `remove_tag` is called).
  - Whitelist enforcement (non-whitelisted fields are ignored).
  - Extensibility of `SENTRY_TAG_FIELDS`.
- **Risk is low:** The changes only affect Sentry tags and do not alter existing run execution behavior, OTel spans, or logging formats.

## Summary by CodeRabbit

* **Documentation**
  * Added architecture guidance on tracing context and Sentry integration, including how to add searchable tracing fields and tag propagation rules.

* **New Features**
  * Tracing context now automatically syncs selected fields to Sentry: non-empty values are stringified and set as tags; cleared values remove corresponding tags to avoid stale data.

* **Tests**
  * Added tests covering tracing-to-Sentry tag propagation and removal behavior.

* **Chores**
  * Updated .gitignore to exclude type-checker cache files.

### DIFF
--- a/engine/trace/span_context.py
+++ b/engine/trace/span_context.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass, field
 from typing import Optional
 from uuid import UUID
 
+import sentry_sdk
 from e2b_code_interpreter import AsyncSandbox
 
 from ada_backend.database.models import CallType, EnvType
@@ -31,6 +32,26 @@ class TracingSpanParams:
 _tracing_context: ContextVar[Optional[TracingSpanParams]] = ContextVar("_tracing_context", default=None)
 
 _SPAN_FIELDS = {f.name for f in dataclasses.fields(TracingSpanParams)}
+SENTRY_TAG_FIELDS = (
+    "cron_id",
+    "trace_id",
+    "project_id",
+    "organization_id",
+    "environment",
+    "call_type",
+    "graph_runner_id",
+    "tag_name",
+)
+
+
+def _sync_to_sentry(params: TracingSpanParams) -> None:
+    isolation_scope = sentry_sdk.get_isolation_scope()
+    for field_name in SENTRY_TAG_FIELDS:
+        field_value = getattr(params, field_name)
+        if field_value is None:
+            isolation_scope.remove_tag(field_name)
+            continue
+        sentry_sdk.set_tag(field_name, str(field_value))
 
 
 def set_tracing_span(**kwargs) -> None:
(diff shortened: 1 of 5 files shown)


engine/trace/span_context.py
```diff
@@ -31,6 +32,26 @@ class TracingSpanParams:
...
+
+
+def _sync_to_sentry(params: TracingSpanParams) -> None:
+    isolation_scope = sentry_sdk.get_isolation_scope()
+    for field_name in SENTRY_TAG_FIELDS:
+        field_value = getattr(params, field_name)
+        if field_value is None:
+            isolation_scope.remove_tag(field_name)
+            continue
+        sentry_sdk.set_tag(field_name, str(field_value))
```
natacha-scopeo, line 54: how trace_id was working before ?

natacha-scopeo (commented):
ok for me but

natacha-scopeo (approved):

engine/trace/span_context.py
pabloscopeo: Yep, normally `trace_id` was already visible in logs before.
Now it’s also sent as a structured Sentry tag, so it's supposed to be easier to filter/search in Sentry.

PR #654, opened 2026-04-15
feat(ci) : Fix Alembic downgrade in staging reset workflow dra-1230

natacha-scopeo:
# Fix Alembic downgrade in staging reset workflow

## Summary
- Fix migration revision parsing that silently skipped downgrades when revision IDs contained non-hex characters (e.g., `g6h7i8j9k0l1`, `hc4u6epu6y03`). The old `grep -oE '^[a-f0-9]+'` pattern matched nothing for these IDs, causing `CURRENT_REV` to be empty and the entire downgrade to be skipped.
- Extract revision IDs from file contents instead of filenames, since some filenames don't match the actual revision declared inside (e.g., `a842d0e56279_add_json_builder.py` declares `revision = "88dcf82ab86"`).
- Replace the fragile one-by-one kubectl exec downgrade loop with a single `alembic downgrade <main_head>` command, computed via set difference of revisions and down_revisions.

## What was broken
The workflow used `grep -oE '^[a-f0-9]+'` to parse Alembic revision IDs, but 5 revisions in the codebase contain characters outside hex range (`g`, `h`, `i`, `k`, `w`, `y`, `z`). The current HEAD of main (`g6h7i8j9k0l1`) is one of them, meaning the downgrade step was effectively a no-op whenever it was needed.

### DIFF
--- a/.github/workflows/reset-k8s-staging-to-main.yml
+++ b/.github/workflows/reset-k8s-staging-to-main.yml
@@ -19,12 +19,46 @@ jobs:
           ref: main
           fetch-depth: 0
 
-      - name: Get main branch migration revisions
+      - name: Get main branch migration revisions and head
         id: migrations
         run: |
-          REVISIONS=$(ls ada_backend/database/alembic/versions/*.py 2>/dev/null | xargs -I {} basename {} .py | cut -d'_' -f1 | tr '\n' ' ')
-          echo "revisions=${REVISIONS}" >> $GITHUB_OUTPUT
-          echo "Main branch revisions: ${REVISIONS}"
+          python3 << 'PYEOF'
+          import ast, os
+
+          versions_dir = "ada_backend/database/alembic/versions"
+          revisions = set()
+          down_revisions = set()
+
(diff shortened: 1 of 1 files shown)


.github/workflows/reset-k8s-staging-to-main.yml
```diff
@@ -19,12 +19,42 @@ jobs:
...
-          echo "Main branch revisions: ${REVISIONS}"
+          python3 << 'PYEOF'
+          import os, re
+
+          versions_dir = "ada_backend/database/alembic/versions"
+          revisions = set()
+          down_revisions = set()
+
+          rev_pattern = re.compile(r"^revision[\s:]+.*?=\s*['\"]([a-zA-Z0-9]+)['\"]", re.MULTILINE)
+          down_pattern = re.compile(r"^down_revision[\s:]+.*?=\s*['\"]([a-zA-Z0-9]+)['\"]", re.MULTILINE)
```
dianascopeo: is there another way to do not regex?

dianascopeo (approved):

PR #655, opened 2026-04-15
feat(components): unify LLM Call output_format with AI Agent dra-1226

natacha-scopeo:
# Unify LLM Call output_format with AI Agent

## Summary
- Align the LLM Call output_format parameter to use the same flat properties format as AI Agent (e.g. `{"answer": {"type": "string"}}`), replacing the previous full OpenAI structured output format
- Enable `drives_output_schema: True` on LLM Call, so setting output_format now creates dynamic output ports that downstream components can reference via field expressions
- Add an Alembic migration to convert existing stored output_format values from the old OpenAI format to flat properties

## Motivation
Until now, LLM Call and AI Agent used different formats for output_format:
- AI Agent: flat properties dict (`drives_output_schema: True`) -- dynamic output ports created
- LLM Call: full OpenAI `json_schema` format (`drives_output_schema: False`) -- no dynamic output ports
This inconsistency meant LLM Call users couldn't wire individual structured output fields to downstream components. With this change, both components accept the same format and both produce dynamic output ports.

## Summary by CodeRabbit

* **New Features**
  * LLM Call output_format now drives dynamic output schema and creates structured response ports for flat property definitions.

* **Bug Fixes**
  * Improved conversion of flat output-format specifications into OpenAI-style structured response format so structured outputs are produced reliably.

* **Documentation**
  * Updated docs to note that certain fields (including LLM Call output_format) affect dynamic output port recreation.

* **Tests**
  * Added/updated tests covering conversion, structured outputs, and dynamic output port behavior.

* **Chores**
  * Migration applied to update stored output-format representations.

### DIFF
--- a/ada_backend/database/alembic/versions/h7i8j9k0l1m2_migrate_llm_call_output_format_to_flat_properties.py
+++ b/ada_backend/database/alembic/versions/h7i8j9k0l1m2_migrate_llm_call_output_format_to_flat_properties.py
@@ -1,0 +1,157 @@
+"""Migrate LLM Call output_format from OpenAI structured output format to flat properties.
+
+Unifies the output_format parameter between LLM Call and AI Agent. Existing
+LLM Call output_format values stored as OpenAI format
+({"name": ..., "schema": {"properties": {...}}}) are rewritten to just the
+flat properties dict ({"answer": {"type": "string"}, ...}).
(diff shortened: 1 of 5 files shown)


dianascopeo (approved):
lgtm

PR #657, opened 2026-04-16
feat(graph): save graph v2 dra-890

natacha-scopeo:
# feat(graphs): file-based v2 graph API — granular component CRUD and topology endpoints DRA-890

## Summary
Introduces a **v2 graph API** with granular, file-based endpoints that allow creating, updating, and deleting individual components without replacing the entire graph. This replaces the monolithic "save everything at once" pattern with a component-level workflow better suited for the front-end and MCP server.

## Release Notes

* **New Features**
  * Added V2 graph API endpoints for granular component and topology management (create, update, delete components; retrieve graph data; update topology).
  * Updated Git sync to support multi-file graph payloads, enabling per-component file organization alongside graph topology files.

* **Documentation**
  * Updated API reference with V2 graph endpoints and request/response schemas.
  * Revised Git sync documentation to reflect multi-file payload structure.
  * Expanded MCP tools documentation to include new V2 graph operations.

* **Tests**
  * Added comprehensive test coverage for V2 graph services and Git sync multi-file behavior.


ada_backend/schemas/pipeline/graph_schema.py
```diff
@@ -76,6 +76,111 @@ class GraphUpdateSchema(BaseModel):
...
+    relationships: list[GraphMapRelationshipSchema] = Field(default_factory=list)
+    last_edited_time: Optional[datetime] = None
+
+
+class GraphGetV2Response(BaseModel):
+    graph_map: GraphMapSchema
+    tag_name: Optional[str] = None
+    change_log: Optional[str] = None
+    last_edited_time: Optional[datetime] = None
+    last_edited_user_id: Optional[UUID] = None
```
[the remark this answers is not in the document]
natacha-scopeo: no they will use the get component endpoint

.cursor/rules/mcp.mdc
```diff
@@ -37,12 +37,13 @@ globs:
...
 - Run retries are run-scoped via `retry_run(project_id, run_id, env|graph_runner_id)` and rely on persisted `run_inputs`; do not reintroduce project-scoped retry policy fields.
 - `update_graph` supports optional optimistic locking: pass `last_edited_time` from a previous response to detect concurrent edits (409 Conflict).
+- File-based graph v2 tools: granular tools (`create_component_v2`, `update_component_v2`, `delete_component_v2`, `update_graph_topology_v2`) for front/MCP. Git sync uses the same service functions internally (no MCP tool). Granular flow: create component → update topology to connect it. Field expression refs in `input_port_instances` support `file_key` (e.g. `{"type": "ref", "file_key": "start", "port": "output"}`) as an alternative to `instance` UUID; the v2 mapper resolves `file_key` → UUID before saving.
```
pabloscopeo, line 47: When looking at the graph.json, I thought that yeah, `file_key` it's a nice name backend-wise, but maybe user-wise it would be clearer to simply put "id". I've seen this in another systems, it's probably more intuitive - even if we know that it's not the true id, for the person/AI writing the json files, it probably makes more sense.

(Out of scope of course, as it makes more sense to do the moment we add structure validation, etc, unless we add a simple file_key -> id alias or something like that)

EDIT: Maybe `ref` or `slug` it's a good name/alias too? if we want to avoid `id` for some reason.

ada_backend/routers/graph_router_v2.py
```diff
@@ -0,0 +1,199 @@
...
+)
+from ada_backend.services.graph.graph_mutation_helpers import (
+    delete_component_v2 as delete_component_v2_service,
+)
+from ada_backend.services.graph.graph_mutation_helpers import (
+    save_graph_topology_v2 as save_graph_topology_v2_service,
+)
+from ada_backend.services.graph.graph_mutation_helpers import (
+    update_component_v2 as update_component_v2_service,
+)
```
pabloscopeo: maybe importing `graph_mutation_helpers` directly it's easier?

pabloscopeo (approved):
LGTM!

.cursor/rules/mcp.mdc
natacha-scopeo: I agree on the idea I will do an iteration

PR #660, opened 2026-04-16
feat(components): fix scorer and categorizer output format dra-1236

natacha-scopeo:
# Fix Scorer and Categorizer output format double-wrapping

## Summary
- Fix a Pydantic validation error (OutputFormatModel) in both the Scorer and Categorizer components caused by their output format dicts being double-wrapped into the OpenAI structured output shape.
- Add regression tests for both components.

## Problem
Both `Scorer.OUTPUT_FORMAT` and `Categorizer._build_output_format()` produced the full OpenAI response format (name, strict, schema with nested properties). This dict was then passed through `_convert_properties_to_openai_format(`), which assumes a flat properties dict and wraps it again. The result was a doubly-nested schema where name (a string) and strict (a bool) appeared as property definitions, causing OutputFormatModel validation to fail.

### DIFF
--- a/engine/components/categorizer.py
+++ b/engine/components/categorizer.py
@@ -57,27 +57,18 @@ def _build_output_format(categories: dict[str, str]) -> str:
     category_names = list(categories.keys())
 
     return json.dumps({
-        "name": "categorization_result",
-        "strict": True,
-        "schema": {
-            "type": "object",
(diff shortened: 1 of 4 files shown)


dianascopeo (approved):
lgtm

PR #662, opened 2026-04-16
feat(Qdrant): Remove check qdrant collection type DRA-1237

dianascopeo:
All Qdrant collections have been migrated to hybrid (dense + sparse vectors) via the migration. 

This PR removes the runtime is_hybrid_collection_async checks that were conditionally switching between dense-only and hybrid behavior

What changed in the Qdrant payload format
After the migration, all collections use named vectors instead of flat vectors:

Search payloads now always include "using": "dense" to target the named dense vector:

Hybrid search uses both named vectors via prefetch "using": "sparse" for BM25 keyword matching and "using": "dense" for semantic

## Summary by CodeRabbit

* **Refactor**
  * Removed runtime hybrid-collection detection and caching; searches now always use dense-mode queries.
  * Stored vectors are standardized to include both dense and sparse fields during ingestion.
  * Migration now reads collection metadata to decide hybrid configuration instead of a separate detection routine.
* **Tests**
  * Removed tests validating hybrid-detection and its caching behavior; relevant tests updated to reflect unconditional behavior.

### DIFF
--- a/engine/qdrant_service.py
+++ b/engine/qdrant_service.py
@@ -157,8 +157,6 @@ class QdrantService:
 
         self.default_schema = default_schema
         self._schemas: dict[str, QdrantCollectionSchema] = {}
-        # TODO Remove when production all collections are hybrid
-        self._hybrid_cache: dict[str, bool] = {}
 
     def register_schema(self, collection_name: str, schema: QdrantCollectionSchema):
         """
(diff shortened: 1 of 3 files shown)


natacha-scopeo (commented):
Do we use hybrid mode all the time now? Shouldn't we use the /rag /retriever parameter to select the mode?

dianascopeo:
> Do we use hybrid mode all the time now? Shouldn't we use the /rag /retriever parameter to select the mode?

there is parameter to choose for new version and old one will alwas use semantic search

natacha-scopeo (approved):
LGTM (need QA all client rag/retriever)

MelMayssonOwen (approved):
LGTM

PR #663, opened 2026-04-16
fix(monorepo): restore files dropped by root gitignore + skip-build o…

MelMayssonOwen:
…utside frontend (DRA-1238)

- Anchor backend-focused rules (/data, /monitoring, /images/*, /.vscode) to root so they don't block frontend subtrees.
- Add !frontend/public/images/** to keep schema PNGs despite the broad *.png ignore.
- Restore lost files from back-office: monitoring component, studio component-definitions, schema PNGs, .vscode workspace config.
- vercel.json: ignoreCommand skips builds when only non-frontend files change (relative to frontend/ as Root Directory).

## Summary by CodeRabbit

* **New Features**
  * Added a monitoring dashboard displaying project metrics and key performance indicators with interactive charts (line, bar, doughnut, radar, and more).
  * Integrated date range and call type filtering for customized monitoring views.
  * Enabled project-specific monitoring insights with conditional visibility based on project selection.


frontend/.vscode/extensions.json
natacha-scopeo, line 1: don't think it is a good idea to version this

frontend/.vscode/settings.json
natacha-scopeo, line 1: idem for this file

natacha-scopeo (approved):

PR #664, opened 2026-04-16
fix(tracing): propagate run_id to sentry tags

pabloscopeo:
## What changed
Fixes Sentry tag propagation by adding `run_id` to the tracing context, preventing `ContextVar` leakage in long-lived workers, and renaming the `environment` tag to `env`.

## Why
Sentry was missing key filters (`run_id`, `project_id`, `org_id`). The root causes were: `run_id` wasn't tracked in the span, worker threads leaked stale context between independent runs, and our `environment` tag shadowed Sentry's native deployment tag.

## Scope / constraints
- Added `reset_tracing_span()` exclusively for top-level worker boundaries.
- Centralized tracing-to-Sentry sync remains in `set_tracing_span`.
- No extra log volume or PII/user-level tagging (`user_id`).

## Validation
Targeted regression suite passed (22 tests):
- `tests/engine/trace/test_span_context.py`
- `tests/ada_backend/workers/test_run_queue_worker.py`

## Release Notes

* **Chores**
  * Enhanced tracing and observability infrastructure to improve debugging capabilities. Implemented stricter isolation of execution context between independent runs to prevent information leakage. Refined metric mapping to external monitoring platforms for better visibility. Added automatic context cleanup at execution boundaries to ensure clean state between operations and improve overall system reliability.

### DIFF
--- a/ada_backend/workers/run_queue_worker.py
+++ b/ada_backend/workers/run_queue_worker.py
@@ -59,9 +59,9 @@ class RunQueueWorker(BaseQueueWorker):
             try:
                 try:
                     env = EnvType(env_str) if env_str else None
-                except ValueError as e:
+                except ValueError:
                     LOGGER.warning("Invalid env in run %s: %s", run_id, env_str)
-                    raise e
+                    raise
 
                 try:
                     call_type = CallType(trigger_str)
--- a/ada_backend/workers/run_queue_worker.py
+++ b/ada_backend/workers/run_queue_worker.py
@@ -14,7 +14,7 @@ from ada_backend.services.run_service import _upload_result_to_s3, update_run_st
 from ada_backend.services.tag_service import compose_tag_name
 from ada_backend.utils.redis_client import publish_run_event
 from ada_backend.workers.base_queue_worker import BaseQueueWorker
-from engine.trace.span_context import set_tracing_span
+from engine.trace.span_context import reset_tracing_span, set_tracing_span
 from settings import settings
 
 LOGGER = logging.getLogger(__name__)
@@ -55,6 +55,7 @@ class RunQueueWorker(BaseQueueWorker):
         trigger_str = payload.get("trigger", CallType.API.value)
 
         with sentry_sdk.isolation_scope():
+            reset_tracing_span()
             set_tracing_span(run_id=str(run_id))
             try:
                 try:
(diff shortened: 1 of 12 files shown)


ada_backend/workers/run_queue_worker.py
```diff
@@ -51,115 +54,118 @@ def process_payload(self, payload: dict, loop: asyncio.AbstractEventLoop) -> Non
...
-
-                now = datetime.now(timezone.utc)
-                update_run_status(
-                    session,
-                    run_id=run_id,
-                    project_id=project_id,
-                    status=RunStatus.RUNNING,
-                    started_at=now,
-                )
+                try:
```
natacha-scopeo, line 61: why try try ?

engine/trace/span_context.py
```diff
@@ -26,32 +26,46 @@ class TracingSpanParams:
...
+        field_value = getattr(params, attr_name)
         if field_value is None:
-            isolation_scope.remove_tag(field_name)
+            isolation_scope.remove_tag(sentry_key)
+            isolation_scope.remove_attribute(sentry_key)
             continue
-        sentry_sdk.set_tag(field_name, str(field_value))
+        str_value = str(field_value)
+        isolation_scope.set_tag(sentry_key, str_value)
+        isolation_scope.set_attribute(sentry_key, str_value)
```
natacha-scopeo, line 68: what is the difference between a tag and a attribute ?

pabloscopeo: Supposedly Sentry treats them as two different channels: tags are more for events/issues/traces, while attributes are for logs. So I set both on purpose, because the whole point here is to make `run_id` / `project_id` / etc filterable in both places.

ada_backend/workers/run_queue_worker.py
pabloscopeo: I did't write that, I just added a `with sentry_sdk.isolation_scope():` which adds indentation, that's why we see it on the diff.
(According to git blame you wrote it btw haha)

engine/graph_runner/graph_runner.py
```diff
@@ -109,7 +109,9 @@ async def run(self, *inputs: AgentPayload | dict, **kwargs) -> AgentPayload | di
             params = get_tracing_span()
             if params:
                 span_json = json.loads(span.to_json())
-                params.trace_id = span_json["context"]["trace_id"]
+                trace_id = span_json["context"]["trace_id"]
+                params.trace_id = trace_id
+                set_tracing_span(trace_id=trace_id)
```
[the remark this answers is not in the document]
pabloscopeo: wrong, look at is_root_execution=True

natacha-scopeo (approved):

PR #667, opened 2026-04-17
feat(traces): Fix trace search with special characters dra-1242

natacha-scopeo:
# Fix trace search with special/accented characters (e.g. é)

## Summary
- Fix empty search results when querying traces with non-ASCII characters (é, ñ, ü, etc.) by storing span message content as native UTF-8 instead of JSON unicode escapes (`\u00e9`)
- Add backward-compatible search that matches both legacy escaped data and new UTF-8 data, plus proper escaping of LIKE metacharacters (`%`, `_`)

## Problem
`json.dumps()` defaults to `ensure_ascii=True`, which stored `"résumé"` as `"r\u00e9sum\u00e9"` in the `span_messages.input_content` column. The `ILIKE '%résumé%'` search looked for the literal UTF-8 character, which never matched the 6-char ASCII escape sequence.

### DIFF
--- a/ada_backend/services/metrics/utils.py
+++ b/ada_backend/services/metrics/utils.py
@@ -92,6 +92,10 @@ def query_trace_duration(
     return df
 
 
+def _escape_ilike(value: str) -> str:
+    return value.replace("!", "!!").replace("%", "!%").replace("_", "!_")
+
+
 def query_root_trace_duration(
     project_id: UUID,
     duration_days: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


ada_backend/services/metrics/utils.py
```diff
@@ -92,6 +92,10 @@ def query_trace_duration(
     return df
 
 
+def _escape_ilike(value: str) -> str:
```
dianascopeo, line 95: what is ilike?

natacha-scopeo: ILIKE is PostgreSQL's case-insensitive pattern matching operator (like LIKE, but ignores case)

dianascopeo (approved):

PR #670, opened 2026-04-20
feat(webhook): use durable redis dra-1239

natacha-scopeo:
# feat(webhook): route internal webhook runs through durable RunQueueWorker (DRA-1239)

## Summary
- **Replace FastAPI background tasks with the durable Redis run queue** for runs triggered via the internal webhook endpoint (`/internal/webhooks/projects/{project_id}/envs/{env}/run`). Runs are now enqueued to the `RunQueueWorker` instead of being executed inline as background tasks, giving them heartbeat-based orphan recovery and consistent execution semantics with all other run paths.
- Move cron lifecycle management into `RunQueueWorker`: the `_execute_run` / `_execute_cron_run` helpers are removed from the router; `CronRun `status transitions (QUEUED→RUNNING→COMPLETED/ERROR) are now handled by the worker via `_finalize_cron_run`, so cron state is updated even if the API process restarts mid-execution.

## Motivation
Previously, the internal webhook endpoint fired runs as FastAPI BackgroundTasks. These tasks had no heartbeat, no orphan recovery, and no visibility in the run queue — if the API pod was killed mid-run the execution was silently lost. Routing through RunQueueWorker unifies the execution path and ensures every run benefits from the existing durability guarantees.

## Summary by CodeRabbit

* **Refactor**
  * Webhook/cron executions now enqueue runs onto a durable Redis-backed queue processed by a worker, enabling heartbeat-based orphan recovery.
  * Cron-triggered runs preserve scheduler tracing and drive explicit status transitions (queued→running→completed/failed).
  * If enqueueing fails the run is marked failed and the API returns HTTP 503.

* **Documentation**
  * Updated webhook and scheduler execution flow to reflect queue-based processing, tracing behavior, and failure handling.


ada_backend/workers/run_queue_worker.py
```diff
@@ -70,10 +94,22 @@ def process_payload(self, payload: dict, loop: asyncio.AbstractEventLoop) -> Non
...
+                            raise
+
                 with get_db_session() as session:
                     run = run_repository.get_run(session, run_id)
                     if not run:
                         LOGGER.warning("Run %s not found, skipping", run_id)
+                        self._finalize_cron_run(cron_run_id, False, f"Run {run_id} not found")
                         return
                     current = run.status if isinstance(run.status, RunStatus) else RunStatus(str(run.status))
                     if current != RunStatus.PENDING:
```
MelMayssonOwen, line 112: you are missing self._finalize_cron_run. it can look in that case

MelMayssonOwen (changes requested):

MelMayssonOwen (approved):

PR #673, opened 2026-04-20
feat(Ingestion): fix chunk_size when we combine chunks together DRA-1251

dianascopeo:
## SUMMARY
chunk size overflow when tree chunker merges chunks with different ancestors

Problem

When _combine_chunks merges two chunks via __add__, ancestor headers get baked into the resulting .content. The token count check only validates the two chunks' content before merging, but __add__ appends ancestor paths into the combined content — producing chunks that exceed the configured chunk_size. This causes downstream embedding failures

Fix
Re-check the combined chunk's token count after __add__, and re-split it with SentenceSplitter if it exceeds the limit.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined chunk combination logic to make document chunking more consistent.
* **Behavior**
  * Changed the default embedding model used for token counting, which may alter chunk sizing and tokenization.
* **Tests**
  * Updated test setup to align with the new default embedding model and chunking behavior.

### DIFF
--- a/data_ingestion/markdown/tree_chunker.py
+++ b/data_ingestion/markdown/tree_chunker.py
@@ -104,8 +104,11 @@ class TreeChunker:
             elif self._count_tokens(current_chunk.content) + self._count_tokens(last_chunk.content) > self._chunk_size:
                 first_chunks.append(last_chunk)
             else:
-                current_chunk += last_chunk
-                first_chunks[-1] = current_chunk
+                combined = current_chunk + last_chunk
+                if self._count_tokens(combined.content) > self._chunk_size:
+                    first_chunks += self._split_text(combined)
+                else:
+                    first_chunks[-1] = combined
             return first_chunks
 
     def _fetch_ancestors_to_level(self, ancestors: list[TreeChunk], level: MarkdownLevel) -> list[TreeChunk]:
--- a/data_ingestion/markdown/tree_chunker.py
+++ b/data_ingestion/markdown/tree_chunker.py
@@ -101,12 +101,10 @@ class TreeChunker:
             last_chunk = chunks[-1]
             if self._count_tokens(last_chunk.content) > self._chunk_size:
                 first_chunks += self._split_text(last_chunk)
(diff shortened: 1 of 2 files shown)


data_ingestion/markdown/tree_chunker.py
```diff
@@ -104,8 +104,11 @@ def _combine_chunks(self, chunks: list):
             elif self._count_tokens(current_chunk.content) + self._count_tokens(last_chunk.content) > self._chunk_size:
                 first_chunks.append(last_chunk)
             else:
-                current_chunk += last_chunk
-                first_chunks[-1] = current_chunk
+                combined = current_chunk + last_chunk
+                if self._count_tokens(combined.content) > self._chunk_size:
+                    first_chunks += self._split_text(combined)
```
natacha-scopeo: why doing this ? I can read line 104 exactly the same check

dianascopeo: Line 104 only checks the .content size of each chunk before merging. But __add__ also injects ancestor headers into the result, so the merged content can be bigger than the sum of the two contents alone

natacha-scopeo: so real fix is to fix line 104 + why you don't do as line 104 ? we are in the same case "combined chunk is too big" and if same case, you do something different.

data_ingestion/markdown/tree_chunker.py
```diff
@@ -101,11 +101,12 @@ def _combine_chunks(self, chunks: list):
             last_chunk = chunks[-1]
             if self._count_tokens(last_chunk.content) > self._chunk_size:
                 first_chunks += self._split_text(last_chunk)
-            elif self._count_tokens(current_chunk.content) + self._count_tokens(last_chunk.content) > self._chunk_size:
-                first_chunks.append(last_chunk)
             else:
-                current_chunk += last_chunk
-                first_chunks[-1] = current_chunk
+                combined = current_chunk + last_chunk
```
natacha-scopeo: combined_chunk

natacha-scopeo (approved):

PR #675, opened 2026-04-20
feat(runs): add org-level runs table dra-1244

natacha-scopeo:
# Add org-level runs table with filters, retry & trace drawer (DRA-1244)

## Summary
- **New org-scoped runs page** (`/org/:orgId/runs`) with a paginated data table showing all runs across projects in the organization, replacing the previous per-project runs endpoint.
- **Rich filtering** — status (multi-select), project (multi-select), trigger type, and date range filters, all wired to backend query params.
- **Trace preview drawer** — click a run's trace to open a slide-out `TracePreviewDrawer` that reuses the existing span tree and detail components.
- **Retry from the table** — retry button per row that fetches stored input and re-enqueues the run, with attempt count displayed.
- **Run input endpoint** — new` GET org/:orgId/runs/:runId/input` to fetch persisted input data for a run, scoped to the org.

## Summary by CodeRabbit

* **New Features**
  * Added organization-scoped run listing endpoint with pagination and optional filtering (statuses, project IDs, trigger, date range).
  * Added endpoint to retrieve run input data for specific runs.
  * Added "Runs" navigation section for browsing organization runs with filtering, sorting, and retry actions.
  * Added trace preview drawer for viewing execution traces.

* **Documentation**
  * Updated API reference for new monitoring endpoints.

* **Chores**
  * Updated MCP Server's `list_runs` tool to support organization context and optional filters.


ada_backend/routers/monitor_router.py
pabloscopeo, line 1: why "monitor"? the front page it's called "monitor" or something like that? if I just read ada_backend/routers/monitor_router.py I wouldn't think this is related to runs

ada_backend/schemas/run_schema.py
```diff
@@ -41,6 +41,13 @@ class RunResponseSchema(BaseModel):
 
     model_config = ConfigDict(from_attributes=True)
 
+    @field_validator("error", mode="before")
+    @classmethod
+    def _coerce_error(cls, v: object) -> object:
+        if isinstance(v, str):
+            return {"message": v}
+        return v
```
pabloscopeo, line 49: what is this for? i'm guessing it wraps the request and puts it inside "message", but why is it necessary?

ada_backend/schemas/run_schema.py
```diff
@@ -51,20 +58,42 @@ class RunListPagination(BaseModel):
...
+    started_at: Optional[datetime] = None
+    finished_at: Optional[datetime] = None
+    created_at: datetime
+
+    @field_validator("error", mode="before")
+    @classmethod
+    def _coerce_error(cls, v: object) -> object:
+        if isinstance(v, str):
+            return {"message": v}
+        return v
```
pabloscopeo, line 89: ah nevermind, I see you just moved the code (pls confirm)

pabloscopeo (approved):
Backend review: LGTM but some doubts regarding router naming.

Also, it would be nice if we could test the repo function as it does a non-trivial query (out of scope, but i think it would be cool if did some setup on the tests, so we can apply some queries on the test DB and then rollback).

ada_backend/schemas/run_schema.py
natacha-scopeo: error is suppose to be a dict but it something a string... this is why we need this wrapper

pabloscopeo: ok thanks!

ada_backend/routers/run_router.py
```diff
@@ -213,3 +186,73 @@ def retry_run_endpoint(
     except Exception as exc:
         LOGGER.error("Failed to retry run %s for project %s", run_id, project_id, exc_info=True)
         raise HTTPException(status_code=500, detail="Unexpected error while retrying run") from exc
+
+
+org_router = APIRouter(prefix="/org", tags=["Runs"])
```
pabloscopeo, line 191: i think you can put "/org/{organization_id}/"

pabloscopeo (approved):
LGTM!

ada_backend/repositories/run_repository.py
```diff
@@ -53,26 +56,107 @@ def get_latest_run_by_retry_group(session: Session, retry_group_id: UUID) -> Opt
...
+        query, statuses=statuses, project_ids=project_ids, trigger=trigger, date_from=date_from, date_to=date_to
+    )
+    return query.count()
 
 
-def get_runs_by_project(
+def get_runs_by_organization(
     session: Session,
-    project_id: UUID,
+    organization_id: UUID,
```
MelMayssonOwen, line 99: isn’t that low level? why use org id at this level?

MelMayssonOwen (approved):
LGTM. this pushes towards a merge of business and client logic data side

PR #677, opened 2026-04-21
feat(sets):  Validate project ownership in get_set_ids_service dra-1111

natacha-scopeo:
# Validate project ownership in `get_set_ids_service`

## Summary
- Adds a project existence and ownership check before listing variable set IDs, rejecting requests with a nonexistent or cross-organization `project_id` as a 404.
- Previously, `list_set_ids_for_project` would silently return org-global set IDs for any UUID — including unrelated or fabricated project IDs — hiding invalid input from callers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened project validation and ownership checks when accessing or updating variables. Requests for non-existent or mismatched projects now return clearer HTTP 404 responses instead of generic errors, improving feedback when retrieving set IDs or upserting organization-scoped variables.

### DIFF
--- a/ada_backend/services/variables_service.py
+++ b/ada_backend/services/variables_service.py
@@ -32,7 +32,13 @@ from ada_backend.schemas.variable_schemas import (
     VariableSetListResponse,
     VariableSetResponse,
 )
-from ada_backend.services.errors import OAuthSetProtectedError, VariableDefinitionNotFound, VariableSetNotFound
+from ada_backend.services.errors import (
+    OAuthSetProtectedError,
+    ProjectNotFound,
+    ProjectNotInOrganization,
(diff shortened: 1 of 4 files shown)


ada_backend/services/variables_service.py
```diff
@@ -163,6 +168,9 @@ def get_set_ids_service(
     organization_id: UUID,
     project_id: UUID,
 ) -> SetIdsResponse:
+    project = get_project(session, project_id=project_id)
+    if not project or project.organization_id != organization_id:
+        raise ProjectNotFound(project_id)
```
dianascopeo: why not to use _validate_project_ids_belong_to_org? doing same thing

dianascopeo (approved):
lgtm

PR #679, opened 2026-04-21
feat(Syncprod/staging): Fix qdrant sync DRA-1258

dianascopeo:
### Problem 
Qdrant sync reported all collections as failed despite successful data migration, due to a _migration_offsets cleanup bug in the migration tool, and collection aliases were not copied to staging.

### Fix
Detect migration success from job logs instead of exit code, clean up _migration_offsets after each job, and sync aliases from prod to staging.


natacha-scopeo (commented):
What was the migration_offset bug ?

dianascopeo:
> What was the migration_offset bug ?

Qdrant-migration tool creates a temporary collection called _migration_offsets to track its progress during data copy. At the end it deletes but as ou qdrant copy ci failed and it did not cleaned up _migration_offsets

natacha-scopeo (approved):

PR #683, opened 2026-04-22
feat(mcp): clarify full-replace semantics for update_component_v2 dra-1265

natacha-scopeo:
# feat(mcp): clarify full-replace semantics for update_component_v2 DRA-1265

## Summary
- Add explicit documentation that `update_component_v2` uses full-replace semantics for parameters and `input_port_instances` — callers must include ALL parameters from `get_graph`, not only the ones being changed; omitted parameters are lost.


mcp_server/tools/graphs.py
```diff
@@ -117,7 +117,13 @@
         description=(
             "Update a single component instance's parameters and ports (v2).\n\n"
             "Only updates the specified component — does not touch topology. "
-            "Optionally update label and is_start_node."
+            "Optionally update label and is_start_node.\n\n"
+            "⚠️ FULL-REPLACE semantics: `parameters` and `input_port_instances` "
```
dianascopeo, line 121: do we need icons in description?

natacha-scopeo: it is not a problem to have icons here. it acts as a visual attention-grabber that makes the LLM more likely to notice

dianascopeo (approved):

PR #691, opened 2026-04-23
feat(graph): fix V2 compt update ignoring input port dra-1271

natacha-scopeo:
# Fix: V2 component update endpoint ignoring input port field expressions

## Summary
- The `PUT /v2/.../components/{instance_id}` endpoint silently dropped input_port_instances — field expressions (e.g. injections in the messages parameter of an AI agent) were never persisted, so the previous value always reappeared after save.
- Added `_sync_input_port_field_expressions` to `component_instance_v2_service.py` which creates, updates, or removes field expressions on input port instances to match the incoming payload.

## Root cause
create_or_update_component_instance processes parameters, integration, and port_configurations, but completely ignores the input_port_instances field on ComponentInstanceSchema. The full graph save path (update_graph_service.py) handles them correctly, but the individual V2 component update path never did.

### DIFF
--- a/ada_backend/services/graph/component_instance_v2_service.py
+++ b/ada_backend/services/graph/component_instance_v2_service.py
@@ -8,10 +8,20 @@ from ada_backend.repositories.component_repository import (
     get_component_instance_by_id,
 )
 from ada_backend.repositories.edge_repository import delete_edge, get_edges
+from ada_backend.repositories.field_expression_repository import (
+    create_field_expression,
+    delete_field_expression_by_id,
+    update_field_expression,
+)
 from ada_backend.repositories.graph_runner_repository import (
     delete_node,
     get_component_nodes,
     upsert_component_node,
+)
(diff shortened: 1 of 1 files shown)


ada_backend/services/graph/component_instance_v2_service.py
```diff
@@ -78,6 +91,58 @@ def create_component_in_graph(
...
+    """
+    db_ports = get_input_port_instances_for_component_instance(
+        session, instance_id, eager_load_field_expression=True,
+    )
+    db_port_by_name: dict[str, object] = {p.name: p for p in db_ports}
+
+    incoming_names: set[str] = set()
+
+    for port_data in incoming_port_instances:
+        fe_data = port_data.get("field_expression")
```
pabloscopeo: i think naming it's fine but maybe you don't haha

pabloscopeo (approved):
LGTM didn't test

PR #692, opened 2026-04-23
feat(ingestion): create EAV source attributes table and dual-write DRA-1272

pabloscopeo:
## What changed

PR1 of the `source_attributes` normalization to EAV (DRA-1168).

- Creates the new `source_attribute_entries` EAV table.
- Backfills existing legacy `source_attributes` rows into EAV rows, preserving JSONB types and explicitly casting `jsonb null` to SQL `NULL`.
- Adds the coexistence SQLAlchemy model `SourceAttributeEntry`.
- Adds dual-write in `create_source()` to both the legacy table and the new EAV table, restricted to a strict `SourceAttributeKey` allowlist.

## Why

The current `source_attributes` table is a wide table where most columns are `NULL` depending on the source type. Normalizing to EAV prevents schema migrations for every new source integration.

This PR is step 1 of a 2-step rollout:
1. **PR1 (this)**: Add table, backfill, dual-write. Reads stay on the legacy table.
2. **PR2 (next)**: Switch reads to EAV, drop legacy table and wide model.

## Out of scope

- Switching reads to the new table.
- Removing the legacy ORM model.
- Expanding persistence beyond the current database-source behavior (Google Drive, Local, and Website sources still do not persist attributes).

## Testing

- Migration explicitly annotated as `migrate-first`.
- Backfill includes embedded sanity checks at the end of the `upgrade()` step to verify row counts and prevent `jsonb null` regressions. Rollback is automatic if checks fail.
- Tested locally with full `db-upgrade` and `db-downgrade` round-trips.

### DIFF
--- a/ada_backend/database/alembic/versions/67df5c87b638_create_source_attribute_entries_table.py
+++ b/ada_backend/database/alembic/versions/67df5c87b638_create_source_attribute_entries_table.py
@@ -1,0 +1,88 @@
+"""create source attribute entries table
+
+Revision ID: 67df5c87b638
+Revises: j1k2l3m4n5o6
+Create Date: 2026-04-23 12:59:35.649247
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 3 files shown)


natacha-scopeo (approved):
LGMT !

PR #693, opened 2026-04-24
feat(runs): Fix missing trace link on failed runs dra-1270

natacha-scopeo:
# Summary
- Failed runs caused by `EngineError` (LLM failures, routing errors, MCP errors, etc.) lost their `trace_id` because the exception was re-raised without capturing it from the tracing span — unlike the generic `Exception` handler which already did this correctly.
- The `run.failed` WebSocket event and the late-connect `get_run_final_state_event` response both omitted `trace_id`, so even when the DB had the trace link, real-time clients never received it. The `run.completed` path already included it.
- The frontend `RunFailedPayload` type, stream callback, and playground error handler had no `trace_id` plumbing, so the UI could never display a trace link for failed runs.

### DIFF
--- a/ada_backend/services/agent_runner_service.py
+++ b/ada_backend/services/agent_runner_service.py
@@ -321,7 +321,9 @@ async def run_agent(
             is_root_execution=True,
         )
         params = get_tracing_span()
-    except EngineError:
+    except EngineError as e:
+        params = get_tracing_span()
+        e.trace_id = params.trace_id if params else None
         raise
     except Exception as e:
         tb = traceback.format_exc()
--- a/frontend/src/composables/useRunStream.ts
+++ b/frontend/src/composables/useRunStream.ts
@@ -22,7 +22,7 @@ export interface RunStreamCallbacks {
     message?: string
     response?: string
   }) => void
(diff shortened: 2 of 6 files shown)


ada_backend/services/agent_runner_service.py
```diff
@@ -321,7 +321,9 @@ async def run_agent(
             is_root_execution=True,
         )
         params = get_tracing_span()
-    except EngineError:
+    except EngineError as e:
+        params = get_tracing_span()
+        e.trace_id = params.trace_id if params else None
```
MelMayssonOwen, line 326: camelcase

frontend/src/composables/useRunStream.ts
```diff
@@ -22,7 +22,7 @@ export interface RunStreamCallbacks {
     message?: string
     response?: string
   }) => void
-  onRunFailed?: (payload: { message: string; type?: string }) => void
+  onRunFailed?: (payload: { message: string; type?: string; trace_id?: string }) => void
```
MelMayssonOwen, line 25: avoid void returns. others are wrong too

MelMayssonOwen (approved):
approved when minor changes made

PR #694, opened 2026-04-24
feat(mcp): fix component update DRA-1274

dianascopeo:
Bug: update_component_parameters silently dropped kind: "input" parameters because they were filtered out before building the API payload.

Fix: Detect kind: "input" parameters and update their field_expressions entries instead of adding them to the regular parameter list, so the V1 graph API correctly persists them.

### DIFF
--- a/mcp_server/tools/graphs.py
+++ b/mcp_server/tools/graphs.py
@@ -536,6 +536,28 @@ def register(mcp: FastMCP) -> None:
                 f"Available parameters: {available}"
             )
 
+        field_expr_list = target.get("field_expressions", [])
+        for param in existing_params:
+            if param.get("kind") != "input" or param.get("name") not in parameters:
+                continue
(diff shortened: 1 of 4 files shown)


mcp_server/tools/graphs.py
```diff
@@ -536,6 +536,31 @@ async def update_component_parameters(
...
+                    existing_fe["expression_text"] = str(new_value)
+                else:
+                    field_expr_list.append({
+                        "field_name": field_name,
+                        "expression_json": expr,
+                        "expression_text": str(new_value),
+                    })
+            elif existing_fe:
+                field_expr_list.remove(existing_fe)
+        target["field_expressions"] = field_expr_list
```
natacha-scopeo, line 562: code duplication ?

natacha-scopeo (approved):

PR #696, opened 2026-04-24
feat(qa): add flexible dataset columns dra-1275

natacha-scopeo:
# feat(qa): add flexible dataset columns with normalized cell storage and dual-write

## Summary
- Introduces a normalized `DatasetCellValue` table (EAV-style) to store per-cell values for QA dataset rows, replacing the implicit `input`/`groundtruth`/`custom_columns` structure with a flexible column-based model.
- Adds `AssociationColumnMapping` to track which dataset columns serve as `input` vs `expected_output` for each project association, auto-populated from system columns when associations are created.
- Implements dual-write: all create/update operations on `InputGroundtruth` now also write to `DatasetCellValue`, and responses include the new `cell_values` field, enabling a gradual migration from the legacy column structure.

## Summary by CodeRabbit

* **New Features**
  * Typed QA columns with roles (Input / Expected Output) and automatic per-association mappings
  * Per-row per-column cell_values in create/update/list responses and APIs; custom columns may include an optional default role
  * Dataset creation now auto-populates system columns for associations

* **Behavior Changes / Bug Fixes**
  * Column listings and CSV import/export use user-only columns by default
  * Deleting a custom column also removes its per-cell data; system columns cannot be modified/deleted

* **Documentation**
  * QA docs updated to describe roles, mappings, and service responsibilities

* **Tests**
  * Added mapping, cascade, and cell_values response tests

### DIFF
--- a/ada_backend/database/alembic/versions/k1l2m3n4o5p6_flexible_qa_datasets.py
+++ b/ada_backend/database/alembic/versions/k1l2m3n4o5p6_flexible_qa_datasets.py
@@ -1,0 +1,283 @@
+"""Add flexible QA dataset columns: DatasetCellValue table, column_role, association_column_mappings.
+
+Migrates existing data into the new normalized cell value table.
+
+Revision ID: k1l2m3n4o5p6
+Revises: b4c5d6e7f8b0
+Create Date: 2026-04-24
+"""
+
+import json
+import uuid
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+from ada_backend.database.utils import create_enum_if_not_exists
+
+revision: str = "k1l2m3n4o5p6"
+down_revision: Union[str, None] = "b4c5d6e7f8b0"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+deploy_strategy = "migrate-first"
+
+QA_SCHEMA = "quality_assurance"
+
+
+def upgrade() -> None:
+    create_enum_if_not_exists(op.get_bind(), ["input", "expected_output"], "column_role", schema=QA_SCHEMA)
+    column_role_enum = postgresql.ENUM(
+        "input", "expected_output", name="column_role", schema=QA_SCHEMA, create_type=False
+    )
+
+    op.add_column(
+        "qa_dataset_metadata",
+        sa.Column("default_role", column_role_enum, nullable=True),
+        schema=QA_SCHEMA,
+    )
+    op.create_unique_constraint("uq_qa_metadata_column_id", "qa_dataset_metadata", ["column_id"], schema=QA_SCHEMA)
+
+    op.create_table(
+        "association_column_mappings",
+        sa.Column("id", sa.UUID(), server_default=sa.text("gen_random_uuid()"), nullable=False),
+        sa.Column(
+            "association_id",
+            sa.UUID(),
+            sa.ForeignKey(f"{QA_SCHEMA}.dataset_project_associations.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column(
+            "column_id",
+            sa.UUID(),
+            sa.ForeignKey(f"{QA_SCHEMA}.qa_dataset_metadata.column_id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column("role", column_role_enum, nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("association_id", "column_id", name="uq_association_column_mapping"),
+        schema=QA_SCHEMA,
+    )
+    op.create_index(
+        op.f("ix_quality_assurance_association_column_mappings_association_id"),
+        "association_column_mappings",
+        ["association_id"],
+        schema=QA_SCHEMA,
+    )
+    op.create_index(
+        op.f("ix_quality_assurance_association_column_mappings_column_id"),
+        "association_column_mappings",
+        ["column_id"],
+        schema=QA_SCHEMA,
+    )
+
+    op.create_table(
+        "dataset_cell_values",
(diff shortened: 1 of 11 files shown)


ada_backend/database/alembic/versions/k1l2m3n4o5p6_flexible_qa_datasets.py
pabloskewes, line 1: for these kind of migration I think it's cool if we can run some simple validations at the end and raise RuntimeError if it ever fails so it's rolledback. Not absolutely necessary but a nice safety net

ada_backend/database/models.py
```diff
@@ -1897,12 +1896,19 @@ class DatasetProjectAssociation(Base):
         nullable=False,
         index=True,
     )
+    input_column_ids = mapped_column(JSONB, nullable=True)
+    output_column_ids = mapped_column(JSONB, nullable=True)
```
pabloskewes: maybe we could normalize using a relationship? otherwise it's hard to operate on here (example: deleting something on the list, or filtering)

pabloskewes (commented):
WIP review

ada_backend/routers/quality_assurance_router.py
```diff
@@ -305,6 +309,8 @@ def rename_column_org_endpoint(
         raise HTTPException(status_code=400, detail=str(e)) from e
     except QAColumnNotFoundError as e:
         raise HTTPException(status_code=404, detail=str(e)) from e
+    except QASystemColumnModificationError as e:
+        raise HTTPException(status_code=400, detail=str(e)) from e
```
pabloscopeo: these shouldn't be necessary after the centralize error handling PR, I don't know what happened here (why it looks it wasn't applied to this router, it seems I left it out on the refactor)

pabloscopeo: yeah I confirmed, I somehow missed it. It's not your PRs concern so you can leave it like this if you want, but you can also remove the except Errors (except for ValueError, which isn't handled automatically for security reasons), the ServiceError handler should handle it and you can reduce boilerplate + reduce drift risk on error messages

ada_backend/database/models.py
```diff
@@ -1928,15 +1934,41 @@ class QADatasetMetadata(Base):
...
+    )
+
+    id = mapped_column(UUID(as_uuid=True), primary_key=True, server_default=func.gen_random_uuid())
+    row_id = mapped_column(
+        UUID(as_uuid=True),
+        ForeignKey("quality_assurance.input_groundtruth.id", ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+    )
+    column_id = mapped_column(UUID(as_uuid=True), nullable=False, index=True)
```
pabloscopeo: Maybe i'm not understanding well (correct me if I'm wrong please) but i thought this was meant to be a FK to `QADatasetMetadata.column_id`? 

Also, why is it related to `column_id` and not to the QADatasetMetadata PK? (talking about the unique constraint defined) If I'm not mistaken, `column_id` it's not unique on its own, intuitively, I would think that a cell it's a FK to a unique row and unique column

Also, there's a check that we probably should do, it's not direct on how to setup as a constraint,  but we probably need to check that `row_id.dataset_id == col_id.dataset_id` right? if it's too annoying to be enforced on DB level we could at least enforce it on the app layer.

pabloscopeo (commented):
Also, maybe it would be cool to put TODOs on the places that will be removed after part 2. (Not necessary of course, but I think it helps with review too)

ada_backend/database/models.py
pabloscopeo: https://github.com/Scopeo/draftnrun/pull/696/changes#r3149005860
still this comment, column_id it's not a FK, why?

ada_backend/database/models.py
```diff
@@ -1910,12 +1964,36 @@ class DatasetProjectAssociation(Base):
...
+    )
+
+    id = mapped_column(UUID(as_uuid=True), primary_key=True, server_default=func.gen_random_uuid())
+    association_id = mapped_column(
+        UUID(as_uuid=True),
+        ForeignKey("quality_assurance.dataset_project_associations.id", ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+    )
+    column_id = mapped_column(UUID(as_uuid=True), nullable=False, index=True)
```
pabloscopeo: why not FK and not linked into QADatasetMetadata PK instead?

ada_backend/database/models.py
pabloscopeo: Even if there is a good reason to use column_id here instead of a FK to QADatasetMetadata, it would still be good to enforce integrity somehow, because right now nothing prevents storing a column_id that does not exist or belongs to a different dataset, (also it doesn't cascade, etc...), but I will not block on this, as you are closer to the system you probably have a good reason for this (but I didn't get it tbh)

ada_backend/schemas/input_groundtruth_schema.py
```diff
@@ -136,12 +144,20 @@ class InputGroundtruthResponse(BaseModel):
...
     custom_columns: Optional[Dict[str, Optional[str]]] = None
+    cell_values: Optional[Dict[str, Optional[str]]] = None
     created_at: datetime
     updated_at: datetime
 
+    @field_validator("cell_values", mode="before")
+    @classmethod
+    def coerce_cell_values(cls, v):
+        if isinstance(v, list):
+            return None
```
pabloscopeo: is it ok to coerce ["something"] to None instead of raising an error?

pabloscopeo (commented):
Overall LGTM

Some doubts on why we use `column_id` (which it's not the PK and i think it's not enforced as unique I think), I would have though that a cell it's a row (PK to inputs...) + a column (PK to QAmetadata...) + data.

ada_backend/database/models.py
natacha-scopeo: because column_id was not he PK. PK was (dataset_id, column_id). I change so now column_id is a PK

pabloscopeo (approved):

PR #699, opened 2026-04-24
feat(graph): Unify input port instances and parameters for API v2 dra-1277

natacha-scopeo:
# Unify input port instances and parameters for API v2

## Summary
- Merge `input_port_instances` into the parameters list on all v2 API write schemas, using kind: `"input" | "parameter"` to distinguish. Input-kind parameters carry a `field_expression` field for their wiring data. The backend v2 service layer splits them internally before persisting to the existing separate database tables.
- Update the MCP server to produce and consume the unified format instead of separate `input_port_instances`, including all documentation and tool descriptions.
- Add `_split_unified_parameters()` helper with comprehensive unit tests, and update all existing tests to use the new format.
- 
## Motivation
The v2 API previously required two separate lists (`parameters `and `input_port_instances`) on every write payload, while the v1 GET response and the frontend already treated them as a single unified parameters array with a kind field. This mismatch added complexity for every consumer of the v2 API. This change aligns the v2 write API with the existing read format, simplifying the contract for the frontend and MCP server.

## Summary by CodeRabbit

* **New Features**
  * Parameters now support optional field expressions, description, port-definition metadata, and an is_tool_input flag; components use a unified parameters array with kind="input" for wiring.
  * Component instances may omit parameters (parameters now optional), and updates preserve existing parameters when parameters are not provided.

* **Behavior Changes**
  * Single-parameter updates now fail when attempting to overwrite non-literal wired inputs; full-replace API must be used to rewire.

* **Documentation**
  * Examples and guidance updated for unified parameter format and revised read/write semantics.

* **Tests**
  * Tests added/updated for splitting, mapping, file-key resolution, syncing, and end-to-end unified-parameter behavior.


ada_backend/services/graph/component_instance_v2_service.py
```diff
@@ -37,14 +38,46 @@
 LOGGER = logging.getLogger(__name__)
 
 
+def _split_unified_parameters(parameters: list[dict]) -> tuple[list[dict], list[InputPortInstanceSchema]]:
```
pabloscopeo: why parameters doesn't have a type?

ada_backend/services/graph/component_instance_v2_service.py
```diff
@@ -37,14 +38,46 @@
...
+    Returns (param_params, input_port_instances) where:
+    - param_params: parameters with kind="parameter" (or missing kind), passed to BasicParameter storage
+    - input_port_instances: parameters with kind="input", converted to InputPortInstanceSchema
+    """
+    param_params: list[dict] = []
+    input_port_instances: list[InputPortInstanceSchema] = []
+
+    for param in parameters:
+        kind = param.get("kind", "parameter")
+        if kind == "input":
```
pabloscopeo: maybe it would be better if this had a pydantic schema and that `kind` was an enum on that schema

ada_backend/services/graph/graph_v2_mapper_service.py
```diff
@@ -74,6 +75,27 @@ def _resolve_input_port_file_keys(
     return resolved
 
 
+def _resolve_unified_params_file_keys(
+    parameters: list[dict],
```
pabloscopeo: same question, why is this untyped?

pabloscopeo (approved):
LGTM. hopefully we can be type-safer

pabloscopeo (approved):
Thanks for using the types

PR #700, opened 2026-04-27
feat(Ingestion): use uuid for chunks for db ingetsion DRA-1173

dianascopeo:
## Summary by CodeRabbit

* **Chores**
  * Converted legacy chunk identifiers to UUIDs in the database and updated vector storage payloads, with migration logging and rollback support.
  * New ingests now assign fresh UUIDs to chunks.
  * Simplified identifier handling in the vector storage service so stored point IDs match source chunk IDs.

* **Tests**
  * Updated tests to use the new raw chunk ID values and validate the revised identifier behavior.


natacha-scopeo (approved):
LGTM (need big qa)

PR #701, opened 2026-04-27
feat(runs): Persist graph_runner_id on runs and simplify retry dra-1278

natacha-scopeo:
# Persist graph_runner_id on runs and simplify retry flow

## Summary
- Add a `graph_runner_id` column to the runs table so each run records the exact graph version it executed against, resolved at execution time by the` RunQueueWorker`.
- Simplify the retry endpoint: retries now reuse the original run's `graph_runner_id` and `env` instead of requiring the caller to specify them. A legacy env fallback is kept for old runs that predate the migration.

## Motivation
Previously, runs only stored `env` (draft/production) but not the specific graph runner used. Since env bindings change over time (deploys, draft edits), there was no way to trace a historical run to its exact graph version. Now graph_runner_id is captured when the worker transitions a run to RUNNING and stored durably on the row.

This also makes retries deterministic: a retry re-executes against the same graph runner as the original run, rather than requiring the caller to pick an env (which may have moved to a different graph version since the original execution).

## Summary by CodeRabbit

* **Improvements**
  * Retry flow simplified: retrying a run no longer requires re-specifying environment/runner; the system reuses the original run’s environment and runner association.
  * Run records now persist an associated runner identifier, improving run tracking and status accuracy.

* **Documentation**
  * API and user-facing docs updated to reflect the simplified retry behavior and persisted runner association.


ada_backend/routers/run_router.py
```diff
@@ -139,16 +139,15 @@ def retry_run_endpoint(
             session=session,
             run_id=run_id,
             project_id=project_id,
-            env=body.env,
-            graph_runner_id=body.graph_runner_id,
+            legacy_env=body.env,
```
dianascopeo, line 142: what is legacy_env?

dianascopeo (approved):
lgtm

PR #702, opened 2026-04-27
feat(ingestion): switch source attribute reads to EAV DRA-1273

pabloscopeo:
## What changed

This PR completes the read-side switch from the legacy wide `source_attributes` table to the EAV `source_attribute_entries` table introduced in DRA-1272.

On the backend side, the legacy `SourceAttributes` ORM model and its related relationships are removed, source creation stops dual-writing to the legacy table, and `get_source_attributes()` now reconstructs the Pydantic schema from EAV rows while still resolving `source_db_url` through `OrganizationSecret`. The coexistence model is also renamed from `SourceAttributeEntry` to `SourceAttribute` to match its long-term role.

The admin view is updated to display the EAV shape (`attribute_name`, `value`) instead of the legacy wide-table shape.

## Why

DRA-1272 created and backfilled the EAV table but deliberately kept reads on the legacy schema. This PR removes that transitional dependency so the application no longer relies on the wide `source_attributes` table at runtime.

For the migration strategy, we intentionally keep the physical EAV table name as `source_attribute_entries` for now and only drop the legacy `source_attributes` table. Renaming the EAV table in the same PR would create a code-first deployment mismatch, because the new code would point at the new name before the migration had run.

## Migration

A new Alembic migration drops the legacy `source_attributes` table with `deploy_strategy = "code-first"`. The downgrade recreates the old table shape so rollback remains possible.

## Out of scope

- Renaming `source_attribute_entries` to `source_attributes` table (because that would require a `breaking` migration)
- Renaming the EAV index/constraint names to match the final table name
- Expanding persisted source attributes beyond the current database-source path

## Validation

- `make db-upgrade` moved Alembic from `k2l3m4n5o6p7` to `93189a98fdf2`
- confirmed `source_attributes` was dropped while `source_attribute_entries` and its indexes/unique constraint remained intact
- `make db-downgrade` restored `source_attributes` and returned Alembic to `k2l3m4n5o6p7`

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal source attribute data handling for improved system efficiency. Enhanced the admin panel to display attribute names and values more clearly alongside existing information.

### DIFF
--- a/ada_backend/admin/admin.py
+++ b/ada_backend/admin/admin.py
@@ -552,7 +552,7 @@ class SourceAdmin(EnhancedModelView, model=db.DataSource):
     ]
 
 
-class SourceAttributesAdmin(EnhancedModelView, model=db.SourceAttributeEntry):
+class SourceAttributesAdmin(EnhancedModelView, model=db.SourceAttribute):
     category = AdminCategory.SOURCES
     icon = "fas fa-database"
     column_list = [
--- a/ada_backend/admin/admin.py
+++ b/ada_backend/admin/admin.py
@@ -552,12 +552,14 @@ class SourceAdmin(EnhancedModelView, model=db.DataSource):
     ]
 
 
-class SourceAttributesAdmin(EnhancedModelView, model=db.SourceAttributes):
+class SourceAttributesAdmin(EnhancedModelView, model=db.SourceAttributeEntry):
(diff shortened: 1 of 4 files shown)


natacha-scopeo (approved):
LGTM ! (need big QA)

PR #706, opened 2026-04-29
feat(graph): Add Org-Level Prompt Library dra-1285

natacha-scopeo:
# Add Org-Level Prompt Library (backend)

## Summary
- Introduces an org-scoped prompt library with independent versioning, subprompt composition (<<section:*>>), and character-level diff between versions
- Adds pin mechanism: eligible input ports (initial_prompt, prompt_template) can be pinned to a specific prompt version, with staleness detection in the graph GET response
- Gradual opt-in: existing workflows are unaffected — ports stay kind=INPUT until explicitly pinned to a library prompt

## Summary by CodeRabbit

* **New Features**
  * Prompt Library: create/manage prompts with versioning, sections, diffs, and organization scoping
  * Pin/unpin prompt versions to component input ports for runtime prompt injection
  * Prompt usage reporting to see where prompts are applied
* **API**
  * New endpoints for prompt CRUD, versioning, diffing, usages, and pin/unpin operations
* **Documentation**
  * Added prompt library and API reference docs
* **Tests**
  * New unit and integration tests covering prompt behavior and diffs


ada_backend/database/models.py
```diff
@@ -2515,3 +2523,66 @@ def github_repo(self) -> str:
         UniqueConstraint("project_id", name="uq_git_sync_project"),
         Index("ix_git_sync_configs_repo_branch", "github_owner", "github_repo_name", "branch"),
     )
+
+
+class Prompt(Base):
```
pabloscopeo: maybe PromptDefinition it's more clearer and more consistent with our existing models?

ada_backend/database/models.py
```diff
@@ -2515,3 +2523,66 @@ def github_repo(self) -> str:
...
+class PromptVersion(Base):
+    __tablename__ = "prompt_versions"
+
+    id = mapped_column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    prompt_id = mapped_column(
+        UUID(as_uuid=True), ForeignKey("prompts.id", ondelete="CASCADE"), nullable=False, index=True
+    )
+    version_number = mapped_column(Integer, nullable=False)
+    content = mapped_column(Text, nullable=False)
+    change_description = mapped_column(Text, nullable=True)
```
pabloscopeo, line 2555: we version the description too, but why not the name in that case? if we change a prompt's name, that would modify Prompt.name directly? I don't mind any of the 2 options, but maybe it would be better if description and name did the same thing? (either versioned here, either not versioned and only last version kept in Prompt), for simplicity?

ada_backend/routers/prompt_router.py
```diff
@@ -0,0 +1,297 @@
...
+            organization_id=item["prompt"].organization_id,
+            name=item["prompt"].name,
+            description=item["prompt"].description,
+            created_by=item["prompt"].created_by,
+            created_at=item["prompt"].created_at,
+            updated_at=item["prompt"].updated_at,
+            latest_version=item["latest_version"],
+        )
+        for item in items
+    ]
```
pabloscopeo: why not wrapped inside the service itself?

ada_backend/routers/prompt_router.py
```diff
@@ -0,0 +1,297 @@
...
+    prompt = create_prompt_service(
+        session=session,
+        organization_id=org_id,
+        name=payload.name,
+        content=payload.content,
+        description=payload.description,
+        sections=payload.sections,
+        created_by=user.id,
+    )
+    session.commit()
```
pabloscopeo: can we do this commit at least on the service? it would be even better to wrap all of this and put it on a service, if we made a CLI we would need to duplicate this code for example

ada_backend/routers/prompt_router.py
```diff
@@ -0,0 +1,297 @@
...
+    user: Annotated[
+        SupabaseUser, Depends(user_has_access_to_organization_dependency(allowed_roles=UserRights.DEVELOPER.value))
+    ],
+    session: Session = Depends(get_db),
+) -> PromptResponseSchema:
+    prompt = update_prompt_metadata_service(session, prompt_id, name=payload.name, description=payload.description)
+    session.commit()
+    latest = get_latest_prompt_version(session, prompt.id)
+    return PromptResponseSchema(
+        id=prompt.id,
```
pabloscopeo: same

pabloscopeo (approved):
WIP Review, LGTM in general

PR #713, opened 2026-05-07
feat(Qdrant): Add text field as index for qdrant filter DRA-1281

dianascopeo:
### Summary
- Add Qdrant text payload index support for human text metadata so filters like Jean Baptiste can match stored values like Jean-Baptiste.
- Keep technical/exact fields (source_id, chunk_id, IDs, URLs, dates, numbers, statuses) unchanged, and explicitly skip payload indexes for chunk body fields (content, chunk).
- Add a database-source-only Alembic migration to recreate eligible existing Qdrant metadata indexes as text, plus docs/rules and regression coverage.

## Summary by CodeRabbit

* **New Features**
  * Improved metadata filtering: identifier fields use exact-match filters; human-readable metadata supports full-text text-indexing and phrase matching.

* **Documentation**
  * Added guidance on payload filtering, index types, and migration behavior for database-ingested sources.

* **Tests**
  * Expanded tests to validate metadata-to-index mapping, payload-schema payloads, and index creation/recreation behavior.

* **Chores**
  * Migration to recreate/adjust existing metadata indexes for consistent filtering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Scopeo/draftnrun/pull/713)


engine/qdrant_service.py
```diff
@@ -25,12 +25,57 @@
...
+    "ascii_folding": True,
+    "phrase_matching": True,
+}
+
+_EXACT_KEYWORD_FIELD_NAMES = {
+    "chunk_id",
+    "source_id",
+    "file_id",
+    "url",
+    "sync_id",
```
natacha-scopeo, line 48: we create an index on sync_id ? what is source_id ?

engine/qdrant_service.py
```diff
@@ -25,12 +25,57 @@
...
+}
+
+_UNINDEXED_PAYLOAD_FIELD_NAMES = {
+    "content",
+    "chunk",  # for tests
```
natacha-scopeo: do we need to keep chunk if it was only for tests ?

engine/qdrant_service.py
```diff
@@ -25,12 +25,57 @@
...
+
+def is_exact_keyword_field(field_name: str) -> bool:
+    normalized_name = field_name.lower()
+    return (
+        normalized_name in _EXACT_KEYWORD_FIELD_NAMES
+        or normalized_name.endswith("_id")
+        or normalized_name.endswith("_uuid")
+        or normalized_name.endswith("_url")
+        or normalized_name == "id"
+        or normalized_name == "uuid"
```
natacha-scopeo: what is this ? the endswith look like hard code

natacha-scopeo: this is source link. it should not be in qdrant service

engine/qdrant_service.py
dianascopeo: no we skip those fields they will stay as keyword

engine/qdrant_service.py
dianascopeo: I can change test to content

engine/qdrant_service.py
dianascopeo: no this is for qdrant because we need to create index in qdrant

dianascopeo: for now we create 2 index keyword and datetime but for some filters we need text

engine/qdrant_service.py
natacha-scopeo: we should keyword index or any index on sync_id

engine/qdrant_service.py
```diff
@@ -25,12 +25,49 @@
...
+_EXACT_KEYWORD_FIELD_NAMES = {
+    "chunk_id",
+    "source_id",
+    "file_id",
+    "url",
+    "sync_id",
+}
+
+_UNINDEXED_PAYLOAD_FIELD_NAMES = {
+    "content",
```
natacha-scopeo, line 53: Why not index on content ?

dianascopeo: we do not need keyword on content because we do not filter content

dianascopeo: and sync_id is only for db inhest sync we do not need keyword for that aswell

engine/qdrant_service.py
```diff
@@ -25,12 +25,49 @@
...
+}
```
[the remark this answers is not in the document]
natacha-scopeo: What is chunk? shouln't be skip too ?

natacha-scopeo (approved):

PR #726, opened 2026-05-29
feat(components): Add optional else routing to If/Else dra-1303

natacha-scopeo:
---
## Summary by cubic
Adds optional Else routing to If/Else via a new enable_false_path parameter. True routes to order=0; false routes to order=1 only when enabled. Otherwise the branch halts (Linear DRA-1303).

- New Features
  - Engine: Adds enable_false_path (default false) and fixed edge orders (0=true, 1=else) using ExecutionDirective SELECTIVE_EDGE_INDICES; when Else is disabled and the condition is true, falls back to CONTINUE to support legacy edges without order. Seeds the parameter, and updates docs to clarify parsing and that booleans belong in component parameters (removed boolean field-expression coercion).
  - Studio: Detects If/Else as a branching node. Renders outputs as ['0'] or ['0','1'] based on enable_false_path, shows a branch icon with “Condition equals True/Else” labels, persists Else as order: 1, and auto-removes Else edges when disabled. Boolean inputs are saved as 'true'/'false'. Tests cover edge orders, Else toggling, and rendering.

- Migration
  - Wire true as order: 0.
  - To use Else, set enable_false_path=true, then wire order: 1.
  - Keep enable_false_path=false (or omit the Else edge) to halt on false. Legacy true edges without order continue to work.

<sup>Written for commit af90bc81ade45d27779886785e856a9245219663. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Scopeo/draftnrun/pull/726?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * If/Else branching: true edges use order 0, false/else edges use order 1; else execution requires enable_false_path (default: false). Studio exposes If/Else routing and conditionally shows else output.

* **Bug Fixes**
  * Skipped subgraph tasks are now marked halted instead of completed. Boolean parameter serialization/parsing and coercion behavior improved for routing/IfElse flows.

* **Documentation**
  * Clarified execution directives, selective-edge semantics, and disabled-else behavior.

* **Tests**
  * Added coverage for If/Else routing, selective-edge indices, halted successors, else-path toggling, and boolean parsing/coercion.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Scopeo/draftnrun/pull/726?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


pabloscopeo (approved):
LGTM, I think we can have the boolean as a param instead of a field expression, that would make the changes in parser.py and coercion_matrix.py not necessary, which would be more ideal: the first one widens the possible inputs for field expressions (which augments complexity) while the second one means that we rely on the coercion matrix for parsing this input, which could be fragile because the coercion matrix was meant to be a runtime glue that we needed to hold everything together due to the fact that we didn't have typed I/O a while ago (so it's a part of the codebase that should ideally be refactored into something more robust, one day).

PR #753, opened 2026-06-29
feat(ingestion): remove supabase storage bucket dra-1326

pabloscopeo:
## Summary

- Remove the Supabase Storage bucket dependency from local knowledge ingestion by keeping uploaded files durably in S3 and deleting the old sync-and-cleanup path.
- Preserve source downloads by storing stable `s3_path` metadata, adding an authenticated S3 download-url endpoint, and updating playground source clicks to use presigned S3 URLs.
- Remove obsolete Supabase bucket configuration/docs and add regression coverage for S3-backed document metadata and download URL generation.

## Out of scope

- Google Drive sources keep using their existing external source URLs; this change only replaces the Supabase bucket path used for local uploads.

---
## Summary by cubic
Removes the Supabase Storage bucket (DRA-1326). Local uploads are stored in S3, and downloads now use presigned S3 URLs resolved on request.

- New Features
  - Added POST `/organizations/{org_id}/files/download-url` (JWT Member) to return a presigned S3 URL; validates the key belongs to the org (400 on invalid) and masks internal storage errors with 500.
  - Knowledge service presigns document links from `metadata.s3_path`; external source URLs (e.g., Google Drive) are unchanged.
  - S3 key sanitization preserves the organization prefix for uploads and downloads; ingestion and S3 folder manager persist `s3_path` in document metadata; `BaseDocument.url` now prefers `source_url` then `s3_path`.
  - Playground/source downloads now call `filesApi.getPresignedDownloadUrl`; no Supabase SDK usage.

- Migration
  - Remove `SUPABASE_BUCKET_NAME` from env and docs; Supabase bucket is no longer used.
  - Delete Supabase uploader and related sync/cleanup paths; S3 is the only durable store.
  - Ensure `S3_BUCKET_NAME` and S3 access settings are configured.

<sup>Written for commit 9bde21d10f32adf2d8bd3fc657369aced733202a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Scopeo/draftnrun/pull/753?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ada_backend/services/s3_files_service.py
```diff
@@ -145,3 +156,19 @@ def generate_s3_upload_presigned_urls_service(
...
+
+
+def generate_s3_download_presigned_url_service(
+    organization_id: UUID,
+    key: str,
+    bucket_name: str = settings.S3_BUCKET_NAME,
+) -> S3DownloadURL:
+    key_prefix = f"{organization_id}/"
+    if not key.startswith(key_prefix):
+        raise S3DownloadKeyValidationError("S3 key does not belong to this organization")
```
natacha-scopeo: leak S3 info in error

ada_backend/services/s3_files_service.py
```diff
@@ -145,3 +156,19 @@ def generate_s3_upload_presigned_urls_service(
             )
         )
     return upload_urls
+
+
+def generate_s3_download_presigned_url_service(
```
natacha-scopeo: this download service is only for ingestion downloaded files ?

data_ingestion/document/folder_management/s3_folder_management.py
```diff
@@ -69,7 +69,10 @@ def _get_file_info(self, file_path: str) -> FileDocument:
             type=file_type,
             file_name=file_name,
             folder_name=str(Path(file_path).parent),
-            metadata=file_data["metadata"],
+            metadata={
+                **file_data["metadata"],
+                "s3_path": file_data["s3_path"],
```
natacha-scopeo, line 74: is this save in Qdrant ? I don't think we should save have  qdrant index for s3_path

ada_backend/services/s3_files_service.py
pabloscopeo: this error is internal, not API-facing, API wise you'd see only "Invalid S3 file key", what do you want me to do in particular in this section?

data_ingestion/document/folder_management/s3_folder_management.py
pabloscopeo: it's stored in qdrant but not indexed (chunk metadata, not vector index)

ada_backend/services/s3_files_service.py
pabloscopeo: yes, this is only for org-scoped knowledge/source files in the ingestion S3 bucket. It follows the same idea as the existing output-file flow: files stay in S3 and the app only gets short-lived presigned URLs. Playground temp/output files still use their separate presign path. Current auth is org membership + org-prefixed key, matching how these keys are generated.

ada_backend/services/s3_files_service.py
natacha-scopeo: something that doesn't say we are using S3

ada_backend/services/s3_files_service.py
natacha-scopeo: can you update the function name ?

natacha-scopeo (approved):

